PR #372, opened 2014-12-06
Allow for icons to be located separately.

minecrafter:
This is especially useful in certain setups (like ours, where we relocate files to separate directories).


samclarke:
Looks good! Thanks for updating.

PR #373, opened 2014-12-11
Set default locales without territory for Portuguese and Italian

rbu:
It is an established best practice to provide location information even when only a client's language can be detected, and not their territory. This is the way the German locale already works, providing 'de' as language key will find the locale. Similarly, for 'en' there is a default 'en' locale (defaulting to en-GB) and two locales for GB and US territory.

However, the following locales have no default currently:
* it (it-IT)
* pt (pt-PT, pt-BR)


brunoais:
Sorry about the delay I lost this notification with time and I just found it. Everything seems fine to me, so I'll merge it.

rbu:
Thanks a lot!

brunoais:
np. And sorry about that.

PR #381, opened 2015-01-12
Update jquery.sceditor.js

### LINKED ISSUE #276: SCEditor does not set proper width / height when initialized hidden
I have tried initializing a textarea within a div that was not visible (display none), due to a collapsible edit form. As the site uses a variable width, so does the SCEditor:
```
        $textarea.sceditor({
...
          "width": "77%",
          "resizeWidth": false
        })
```

Unfortunately, while the outer width (on sceditor-container) is applied correctly, the inner width (on the wysiwygEditor iframe) is set to an incorrect fixed value. This is due to a code path that tries to determine a pixel width from a relative width and which fails because jQuery cannot be relied upon to return any meaningful value when asking for the width of an invisible element,
SCEditor really should avoid calculating widths when initalized with non-integer values.

joechilds:
Fixes #276. Looks like it wants to go into lib/SCEditor.js, initEditor() function (approx line 410) :-)

### DIFF
--- a/src/jquery.sceditor.js
+++ b/src/jquery.sceditor.js
@@ -450,7 +450,7 @@
 		 * @private
 		 */
 		initEditor = function () {
-			var doc, tabIndex;
+			var doc, tabIndex, $clone, width, height;
 
 			$sourceEditor  = $('<textarea></textarea>').hide();
 			$wysiwygEditor = $('<iframe frameborder="0"></iframe>');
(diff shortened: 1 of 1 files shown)


brunoais:
Please rebase on this branch and try again ;)

joechilds:
Sorry, I've already spent some of my work day contributing :-) I've got a working copy, just giving something back with the time I have spare heh.

brunoais:
There's no hurry, don't worry. Sam also takes a while to come and do stuff here...

brunoais:
Too long without any answers. Closing...

PR #382, opened 2015-01-16
Closing balises

Pitouli:
bug correction: it was "<span>" and "<div>" instead of "</span>" and "</div>" to close underline / strike / left / center / right / justify


brunoais:
Please push to 1.4.6-dev

PR #384, opened 2015-01-18
Added signal/event to handle pasted data

francisschiavo:
I added a new event (signalPasteData) to allow custom plugins to handle pasted data.
The event is fired only if enablePasteFiltering is set to true.

There is also an optional plugin to disable image streams in img src attributes.

My motivation for those changes are to provide a way to handle pasted data and filter some contents, e.g. disable base64 embedded images on bbcode.


brunoais:
Please:

1. Do the PR towards 1.4.6-dev
2. Do tests towards your changes in the PR. To make sure everything is correct.
3. Make more explicit comments on what that plugin actually does. I can't quite understand it.

PR #414, opened 2015-04-04
Add a way of making and controlling as code inside an element/tag

brunoais:


brunoais:
@samclarke
Please review this PR and confirm if all looks correct according to the ideology of SCE.

brunoais:
I don't like the line breaks I used but I can't think of any better one due to how jshint is forcing me. The 80 char lines is very limiting and is making the code harder to read.

samclarke:
I think might be better as a plugin as it's quite big and not everyone needs code support. What do you think?

I've just done a quick test and when pressing enter twice it seems to cause text to be removed in WYSWYG mode when the caret isn't in a code tag.

Also IE <= 8 doesn't support DOM Ranges or `textContent` so it would need to be updated to support TextRanges. Getting TextRanges to work the same way can be a pain though so it might be easier to just insert markers are the start/end of the selection and work from that. Something like: http://jsfiddle.net/9qbsxnor/

brunoais:
I don't mind making it a plugin. I just don't know how to do it properly with the current call points. I don't know how many would like such thing, too.
Still, some commits or parts of commits are well useful, for example, the use of `isPreFormatted` more extensively when a tag has it.

I can't reproduce the bug you mention. In which browser is it? Is it a collapsed selection? Can you make a video of it?

Oh... Right... IE8... This isn't an essential feature and I don't make code using deprecated or obselete stuff. IE8 is deprecated by microsoft. Feel free to translate it but I won't... Too much work to learn a dead-end API.

As for that code, here is the same code but with my comments on what you did http://jsfiddle.net/9qbsxnor/1/ . It is quite good but I wonder about some decisions made. Anyhow, you were able to use the markers in a way that I was unable to, that's quite nice. It just seems I wasn't using them properly, then. Check my comments about missing features too, please.

samclarke:
It happens in both Chrome and FF with collapsed selection. [Here is a video](https://dl.dropboxusercontent.com/u/56830471/newlinebug.ogv).

This adds API methods which need to work in IE8 which is the problem and IE 8 needs to be supported until its usage drops low enough. The fact it's deprecated doesn't seem to make much difference sadly. Microsoft stopped supporting XP last year and it's still more popular than Mac and Linux combined which is a bit depressing.

When IE 8 usage does finally drop it will allow quite a lot of code to be removed which will be nice but until then it has to remain. If this was a plugin then IE 8 could be ignored as this wouldn't be in the core of the editor.

The JSFiddle was really only meant as an example of a way that should work in IE without messing with TextRanges. It would need tidying and those missing bits added to be used.

src/lib/SCEditor.js
```diff
@@ -1414,6 +1430,200 @@ define(function (require) {
...
+				e.which !== 9		// tab
+			) {
+				return;
+			}
+
+			if (e.shiftKey && e.which !== 9) {
+				return;
+			}
+
+			rangeHelper.mergeTextNodesAtCaret();
```
samclarke: Could this be what's causing the text to disappear? The `mergeTextNodesAtCaret()` method says it causes HTML to be removed.

src/lib/SCEditor.js
```diff
@@ -1414,6 +1430,200 @@ define(function (require) {
...
+
+			var currentRange = rangeHelper.selectedRange();
+
+			if (!rangeHelper.getFirstBlockParent().
+				matches(e.data.searchSelector)) {
```
samclarke: Just noticed `matches()`  needs to be prefixed in IE (yes, IE really is a pain) and also for older Android and iOS: http://caniuse.com/#feat=matchesselector

It might be easier to use the jQuery `is()` instead since that should work everywhere.

brunoais:
Checkout the changes to RangeHelper.getOuterText I made. Reproducing that bug isn't that easy. It requires some typing, clearing and retyping. Sounds like it requires broken text nodes and with other nodes mixed in... It is part of what I tried to do about it messing up when I tried using the markers...

IE8 is at 2.1%, according to StatCounter... I don't understand why so much. Anyway, I don't think that 2.1% is enough for making it special for requiring support for non-essential features.
Are you waiting for the 1% magic number :)?

Ah, OK. got it... I misunderstood it :S.

src/lib/SCEditor.js
brunoais: I don't think so. That code is only executed if all the above is false. These events are associated with certain HTML tags. In this case, this method must not be called as the selector (2nd parameter in the jQuery's .on()) is used.

src/lib/SCEditor.js
brunoais: Oh right... I forgot... yeah... That could be used.

samclarke:
To reproduce it should just be type then hit enter twice, at least that's what does it for me. I didn't clear anything in that video, that was the text disappearing.

Any API methods need to work in all supported browsers. 2.1% is higher than IE9, IE10, Safari and Opera which is just too high for now. It also depends on which stats you use, some are even higher! :( Part of me really wants to ditch support for IE 8 right now but as long as it's usage is so high it needs to stay.

If this is a plugin then it doesn't matter as it doesn't modify the API and can support as few browsers as makes sense.

src/lib/SCEditor.js
samclarke: I've just tested it again and it is being called outside of code blocks. It runs until it reaches the `matches()` test so maybe if that was moved above `mergeTextNodesAtCaret()` it might fix it.

brunoais:
Yeah, I know... Meanwhile I tested on Chrome and I was able to reproduce just like you had shown.

src/lib/RangeHelper.js
```diff
@@ -538,6 +538,57 @@ define(function (require) {
...
+
+			while (currentElem) {
+				fullOffset += currentElem.length;
+				currentElem = currentElem.previousSibling;
+			}
+
+			fullOffset += currentRange.startOffset;
+
+			var text = currentTextNode.wholeText;
+			currentTextNode.parentNode.textContent = text;
```
brunoais: Should I change this to a call to `.normalize()`?

PR #417, opened 2015-04-06
French localization and prettier shortcuts

KaseiFR:
Hi,

I modified some UI strings for a school project, here they are if you're interested.

First time using npm and Grunt, nice tools! The Jscs tests failed though, because of some line breaks in parts of the code I didn't touch.


brunoais:
Please separate the translation and the shortcut prettification into two different PR.

PR #419, opened 2015-04-06
Prettier shortcut

KaseiFR:
See PR #417


samclarke:
Looks good, thanks for the PR!

PR #420, opened 2015-04-07
Remove e.stopPropagation(), use target check in handleDocumentClick instead

locpd:
I wanted to bind to click event in a dropdown from outside code, but couldn't do it because of e.stopPropagation() in click's callback function. By removing it, and add check to handleDocumentClick to check if click event target is inside a dropdown, we can bind to click events in dropdown normally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ sceditor.sublime-workspace
 /dist
 /minified
 /docs
+/nbproject/
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ sceditor.sublime-workspace
 /dist
 /minified
 /docs
-/nbproject/
--- a/src/lib/SCEditor.js
+++ b/src/lib/SCEditor.js
@@ -1400,11 +1400,7 @@ define(function (require) {
 			$dropdown = $('<div class="sceditor-dropdown ' + cssClass + '" />')
 				.css(dropDownCss)
 				.append(content)
-				.appendTo($('body'))
-				.on('click focusin', function (e) {
-					// stop clicks within the dropdown from being handled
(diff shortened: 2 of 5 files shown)


brunoais:
Why do you want to capture clicks outside the dropdown?
If you want that much to capture clicks, why not placing a mock `<div>` inside it?

locpd:
@brunoais I created a custom dropdown for the image tool. Inside this dropdown, user can either paste an image url to insert the image into the editor, or they can click a link which will open a bootstrap dialog  (http://nakupanda.github.io/bootstrap3-dialog/) with upload functionalities so that they can upload image from their pc. I assign an id to this link, and bind a function to clicks on this id from *outside* code, because I don't want the dropdown code to depend on bootstrap.

brunoais:
I didn't fully understand what you want but, from what I understood, you are doing it wrong...
Anyway, just add a containing block element inside the dropdown inside that thing's popin and that should work for you. I don't think that's too hard to accomplish.

Anyhow, I think you are doing it wrong by using that thing to accomplish what you want.

locpd:
@brunoais I just reviewed my code and realized I got it wrong, the issue was specific to how I structure my code only. So this PR should be rejected.
Btw, I added a commit regarding https for youtube tag, which I think is useful, and it shows up here. Do I need to create a new PR for this commit? I'm new to github...

brunoais:
> Btw, I added a commit regarding https for youtube tag, which I think is useful, and it shows up here. Do I need to create a new PR for this commit? I'm new to github...
Yes you'd do.
Anyway, it should be HTTP for an HTTP page and HTTPS for an HTTPS.
Do that in a single commit.

PR #441, opened 2015-05-13
Give a nice API to allow deleting HTML Elements inside selection

brunoais:
@samclarke Please confirm if this update follows the ideology of SCE. This would be very useful for me in how I plan to use SCE.


brunoais:
@samclarke After you are done with the current active issue, please take a look at this. This is a very useful feature for me

src/lib/RangeHelper.js
```diff
@@ -386,6 +386,127 @@ define(function (require) {
...
+		 * @param {DOMElement} commonParent The common ancestor at clipping
+		 * @param {DOMNode} cutElement Where the closing and re-opening happens
+		 * @return undefined
+		 * @function
+		 * @name splitElement
+		 * @memberOf RangeHelper.prototype
+		 * @author http://stackoverflow.com/a/27498178/551625
+		 * @author brunoais
+		 */
+		base.splitElement = function (commonParent, cutElement) {
```
samclarke: Looks good. Could you move this method to the dom module?

src/lib/RangeHelper.js
```diff
@@ -386,6 +386,127 @@ define(function (require) {
...
+
+			var parentOfRemoval = startMarkerElement.nextElementSibling;
+
+			while (parentOfRemoval && parentOfRemoval.id !== endMarker) {
+
+				var removalElement = ($(parentOfRemoval).
+					is(removeElementSelector) && parentOfRemoval) || null;
+
+				var removalElements = parentOfRemoval.
+					querySelectorAll(removeElementSelector);
```
samclarke: Could you change this to use jQuery instead of `querySelectorAll` that way we can use jQuery selectors and it's also consistent.

src/lib/RangeHelper.js
```diff
@@ -386,6 +386,127 @@ define(function (require) {
...
+				for (var i = 0; i < removalElements.length; i++) {
+					fragment = document.createDocumentFragment();
+					while (removalElements[i].firstChild) {
+						fragment.appendChild(removalElements[i].firstChild);
+					}
+					removalElements[i].parentNode.replaceChild(fragment,
+						removalElements[i]);
+				}
+
+				parentOfRemoval = parentOfRemoval.nextElementSibling;
```
samclarke: IE 8 doesn't support the `nextElementSibling` property (IE < 9 is very crappy), could you use the jQuery `$(parentOfRemoval).next()` method instead?

src/lib/RangeHelper.js
brunoais: I thought RangeHelper was meant to run with DOM directly and not jQuery. OK.

brunoais: OK.

src/lib/RangeHelper.js
brunoais: https://api.jquery.com/next/
> Given a jQuery object that represents a set of DOM elements, the .next() method allows us to search through the immediately following sibling of these elements in the DOM tree and construct a new jQuery object from the matching elements.

Answer: No. According to that quote from the manual, `.next()` returns a DOMNode wrapped in jQuery which may or may not be a DOMElement. What I need is the next DOMElement. I'll see if I find an alternative method.

src/lib/RangeHelper.js
brunoais: Sure.

src/lib/RangeHelper.js
samclarke: >Given a jQuery object that represents a set of DOM elements, the .next() method allows us to search through the immediately following sibling of these elements in the DOM tree and construct a new jQuery object from the matching elements.

As `next()` is only being called on one element it will just return a jQuery object containing only the next element. It also only matches elements so it will give the same result `nextElementSibling`.

src/lib/RangeHelper.js
```diff
@@ -386,6 +386,127 @@ define(function (require) {
...
+					is(removeElementSelector) && parentOfRemoval) || null;
+
+				var removalElements = parentOfRemoval.
+					querySelectorAll(removeElementSelector);
+
+				var fragment;
+				// This has to be after querySelectorAll to allow specifying a
+				// selector for multiple HTML tags forming the removal
+				if (removalElement) {
+					fragment = document.createDocumentFragment();
```
samclarke: I've just noticed, as we're using jQuery, this could be simplified to: $(removalElement).replaceWith(removalElement.childNodes);

src/lib/RangeHelper.js
```diff
@@ -386,6 +386,127 @@ define(function (require) {
...
+		 */
+		base.splitAtSelection = function (removeElementSelector) {
+
+			var commonParent = base.getFirstBlockParent();
+
+			base.insertMarkers();
+
+			var startMarkerElement = base.getMarker(startMarker);
+			var endMarkerElement = base.getMarker(endMarker);
+			startMarkerElement.parentNode.normalize();
```
samclarke: Is `normalize()` needed here? Just it can mess up the selection in IE9+ (not always though, it's a pain to reproduce it). Luckily it can be fixed by calling `restoreSelection()` instead of `removeMarkers()` so it isn't too bad if it is needed.

src/lib/RangeHelper.js
```diff
@@ -385,6 +385,82 @@ define(function (require) {
...
+		 *
+		 *
+		 * @param {String} removeElementSelector Selector match what to delete
+		 * @return undefined
+		 * @function
+		 * @name splitAtSelection
+		 * @memberOf RangeHelper.prototype
+		 * @author brunoais
+		 */
+		base.splitAtSelection = function (removeElementSelector) {
```
samclarke: Thinking about it, being able to remove elements only from the selection could be really useful! It might be worth splitting this into two methods, `splitAtSelection()` and `removeFromSelection()` (or some better name if you can think of it).

src/lib/RangeHelper.js
brunoais: Uh... I'm not sure on how this method works. Is it just replacing that 6 -liner excerpt with that code?

src/lib/RangeHelper.js
brunoais: The issue is that this has to be made in the same "insertMarkers session" to avoid edge-case issues. How do you suggest I do that? A wrapper method for those 2 or something of sorts?

src/lib/RangeHelper.js
brunoais: This one is needed because the markers screw up positioning, at least, in firefox. If I don't normalize, I get weird cases of the end marker before the start marker (at rendering time) or uncalled empty text nodes between the start marker and the end marker when there shouldn't be any.

src/lib/RangeHelper.js
samclarke: What are the edge case issues? I'm just wondering if they could be worked around as having the ability to remove nodes from the selection could be very useful.

src/lib/RangeHelper.js
samclarke: In that case `removeMarkers()` will need to be replace with `restoreRange()` to reset the range so IE 9+ doesn't mess up.

> If I don't normalize, I get weird cases of the end marker before the start marker (at rendering time) or uncalled empty text nodes between the start marker and the end marker when there shouldn't be any.

That should really be fixed. Do you have a way to reproduce it?

src/lib/RangeHelper.js
samclarke: Yes. It will replace the `removalElement` element with the array of `childNodes`.

src/lib/RangeHelper.js
brunoais: The one I could reproduce best was using firefox 37.0.2 with a collapsed selection, after selection is restored after the split, the bold inline tag was being removed until the end of the parent block tag instead of just at collapsed selection.
Some others were hard to reproduce and I don't even know if they can be simplified.

src/lib/RangeHelper.js
brunoais: To reproduce, Use multi-level combination of inline elements then do a collapsed selection followed by calling this method with an empty selector. Don't forget to add a return before the `.removeMarkers()` (later replaced by `.restoreSelection()`) so that you can see where the markers are.

brunoais: @samclarke I can't find a `.restoreSelection()` method.

brunoais:
@samclarke Anything else in mind preventing this?

PR #445, opened 2015-05-18
Fixes #444. Running in source mode only was not checking properly

### LINKED ISSUE #444: No proper checks are being made to load SCE into source mode only
If `runWithoutWysiwygSupport` or `isWysiwygSupported` states that WYSIWYG is not supported, it shouldn't assume that both modes are not supported.
Instead, it should lockdown on source mode and not allow switching to WYSIWYG mode.

brunoais:
@samclarke Please check on this.

### DIFF
--- a/src/jquery.sceditor.js
+++ b/src/jquery.sceditor.js
@@ -69,8 +69,8 @@ define(function (require) {
 
 		options = options || {};
 
-		if (!options.runWithoutWysiwygSupport && !browser.isWysiwygSupported) {
-			return;
+		if (options.runWithoutWysiwygSupport || !browser.isWysiwygSupported) {
+			SCEditor.isWysiwygSupported = false;
 		}
 
 		this.each(function () {
--- a/src/lib/SCEditor.js
+++ b/src/lib/SCEditor.js
@@ -341,7 +341,7 @@ define(function (require) {
 
 			// force into source mode if is a browser that can't handle
 			// full editing
-			if (!browser.isWysiwygSupported) {
+			if (!SCEditor.isWysiwygSupported) {
 				base.toggleSourceMode();
 			}
 
@@ -2234,7 +2234,7 @@ define(function (require) {
 			var sourceMode = base.inSourceMode();
 
 			// don't allow switching to WYSIWYG if doesn't support it
-			if (!browser.isWysiwygSupported && sourceMode) {
+			if (!SCEditor.isWysiwygSupported && sourceMode) {
 				return;
 			}
 


src/jquery.sceditor.js
```diff
@@ -69,8 +69,8 @@ define(function (require) {
 
 		options = options || {};
 
-		if (!options.runWithoutWysiwygSupport && !browser.isWysiwygSupported) {
-			return;
+		if (options.runWithoutWysiwygSupport || !browser.isWysiwygSupported) {
+			SCEditor.isWysiwygSupported = false;
```
samclarke: If `runWithoutWysiwygSupport` is set to `false` and `isWysiwygSupported` is set to `false` the editor will still run which is incorrect. It should only run without WYSIWYG support when `runWithoutWysiwygSupport` is `true`.

brunoais: I misinterpreted it, then...
Isn't that an option to force it to run without wysiwyg support?

samclarke: If it's `true` it should force the editor to run in source mode without WYSIWYG support. If it's `false` the editor should be disabled.

With `(!options.runWithoutWysiwygSupport && !browser.isWysiwygSupported)` both WYSIYWG support and `runWithoutWysiwygSupport` have to be false for it to not run. If the editor runs and `isWysiwygSupported` is `false` then it locks into source mode.

brunoais: That's weird... It means that that option between `true` and `false` does two things and that there's no middle ground.
What if I want to force it to run in source mode regardless of the situation?

samclarke: Oh, do you mean force source mode only regardless of WYSIWYG support?

The `runWithoutWysiwygSupport` option is to make the editor run when there is no browser WYSIWYG support. This is what it should do:

| WYSIWYG supported | Run without WYSIWYG |       Mode       |
|:-----------------:|:-------------------:|:----------------:|
|        Yes        |         Yes         |        Any       |
|        Yes        |          No         |        Any       |
|         No        |         Yes         | Source mode only |
|         No        |          No         |     Disabled     |

Forcing the editor to only run in source mode, regardless of if the browser supports WYSIWYG, isn't what `runWithoutWysiwygSupport` option is for. `runWithoutWysiwygSupport` is for what to do if there's no browser WYSIWYG support.

brunoais: OK, got it. In that case, This Pr is moot and a new feature should be made for it.

Issue #422, opened 2015-04-07
Option to start in source mode


OlafvdSpek:
Could an option be added to start the editor in source mode?

brunoais:
I think so.

Destroy666x:
That's already available.
`editorvar.sourceMode(true);`

brunoais:
That doesn't start it in source mode :). That sets the mode.

brunoais:
Solved through #446.

PR #446, opened 2015-05-18
Allow starting editor in source mode

brunoais:
Requires additional changes before being complete.
~~Depends on #444 as it requires making an extra change to it.~~

After closing; solves: #422

### DIFF
--- a/src/lib/SCEditor.js
+++ b/src/lib/SCEditor.js
@@ -334,7 +334,7 @@ define(function (require) {
 			initPlugins();
 			initEmoticons();
 			initToolBar();
-			initEditor();
+			initEditor(!!options.startInSourceMode);
 			initCommands();
 			initOptions();
 			initEvents();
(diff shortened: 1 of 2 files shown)


brunoais:
@samclarke Please comment on this if it looks correct.

src/lib/SCEditor.js
```diff
@@ -404,16 +404,30 @@ define(function (require) {
...
+		initEditor = function (startInSourceMode) {
 			var doc, tabIndex;
 
-			$sourceEditor  = $('<textarea></textarea>').hide();
+			$sourceEditor  = $('<textarea></textarea>');
 			$wysiwygEditor = $(
 				'<iframe frameborder="0" allowfullscreen="true"></iframe>'
 			);
 
+			/* This needs to be done right after they are created because,
```
samclarke: There's a [check for](https://github.com/samclarke/SCEditor/blob/master/src/lib/SCEditor.js#L344) `isWysiwygSupported` in the `init()` method. Changing that to include the `startInSourceMode` option should save a few bytes as it's just:

    if (!browser.isWysiwygSupported || options.startInSourceMode) {

brunoais: OK. I'll look into it.

brunoais: @samclarke That doesn't work. If I do that, it will do BBCode to HTML (BBCode is an example there) transformation followed by HTML to BBCode transformation.
I don't want that to happen. I want it to actually start in source code mode and that source to be the exact content that came from the server

samclarke:
> I don't want that to happen. I want it to actually start in source code mode and that source to be the exact content that came from the server

In that case looks good! Merged.

PR #465, opened 2015-07-30
Fix bug with ie version classes set incorrectly

prontiol:


brunoais:
What was it doing wrong?

Issue #469, opened 2015-08-07
Youtube iframe rendering fails when on https page


ed-sullivation:
This is browser dependent, but would probably be a best practice to lead with "//" instead of "http://" when recreating the embed url to make it protocol independent.

PR #490, opened 2015-11-10
Use https schema for YouTube URLs

### LINKED ISSUE #489: Rendering of youtube links does not work on HTTPS sites
When serving an instance of SCEditor from an HTTPS-protected site, modern browsers require inline content such as iframes are also served via HTTPS. Otherwise, mixed content warnings or errors occur.

This is most notable with the youtube plugin which renders inline content. Since youtube redirects http traffic to https anyway, ```https://``` can simply be used. Otherwise, protocol-relative urls ```///www.youtube.com``` could be considered.

See attached screen shot and PR.

sceditor-youtube-https

rbu:
Fixes #489

### DIFF
--- a/src/lib/templates.js
+++ b/src/lib/templates.js
@@ -75,13 +75,13 @@ define(function () {
 
 		youtubeMenu:
 			'<div><label for="link">{label}</label> ' +
-				'<input type="text" id="link" placeholder="http://" /></div>' +
+				'<input type="text" id="link" placeholder="https://" /></div>' +
 			'<div><input type="button" class="button" value="{insert}" />' +
 				'</div>',
 
 		youtube:
 			'<iframe width="560" height="315" ' +
-			'src="//www.youtube.com/embed/{id}?wmode=opaque" ' +
+			'src="https://www.youtube.com/embed/{id}?wmode=opaque" ' +
 			'data-youtube-id="{id}" frameborder="0" allowfullscreen></iframe>'
 	};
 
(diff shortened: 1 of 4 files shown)


brunoais:
Please compare with:
https://github.com/samclarke/SCEditor/commit/4ddf9a4bcd8d9e604850c29ae220630ef453a6db

rbu:
@brunoais, thank you for the quick feedback. I have to admit, I don't understand the branching concept. It looked like master is where pull requests are being merged, as 1.4.6 is "behind".

In relation to 4ddf9a4 and issue #469, that fix was incomplete as xhtml and bbedit editor templates were not touched. I also touched the same line that commit changed as well, moving from protocol-relative youtube links to explicit https.
YouTube redirects all traffic from http to https anyway, so there's no point in even generating an iframe loading content from http://www.youtube.com/... because that's just a useless redirect.

brunoais:
Sorry. I forgot he changed how it works. This was the right branch.

brunoais:
I wonder if you should instead just the move the `http` instead of making it `https`.
Did you try doing like that?

rbu:
Do you mean rendering ```<iframe src="//www.youtube``` instead of ```<iframe src="https://www.youtube```?

I haven't tried that, but it would solve the problem as well. The reason I chose to force https instead of letting the browser decide based on where SCEditor is rendered is that you cannot watch YouTube without HTTPS. Try opening http://www.youtube.com/ in your browser and check the location bar afterwards.

Destroy666x:
Yeah, `https://` is better than protocol-relative URL here.

brunoais:
OK. 2 Agree, I see nothing against it.

PR #491, opened 2015-11-11
Use https schema for YouTube URLs

### LINKED ISSUE #489: Rendering of youtube links does not work on HTTPS sites

rbu:
Fixes #489

cherry-pick of #490 to 1.4.6-dev

### DIFF
--- a/src/lib/templates.js
+++ b/src/lib/templates.js
@@ -75,13 +75,13 @@ define(function () {
 
 		youtubeMenu:
 			'<div><label for="link">{label}</label> ' +
-				'<input type="text" id="link" placeholder="http://" /></div>' +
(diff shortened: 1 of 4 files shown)


brunoais:
Sorry my mistake.

PR #508, opened 2016-02-24
Optimize PNG assets

rbu:
Applied a lossless PNG recompression using optipng and advancecomp.
Saved a total of 14461 bytes.


samclarke:
Looks good! Merged.

PR #533, opened 2016-08-26
Improve the filtering of emtpy nodes (xhtml plugin)

kt3k:
xhtml plugin filters the following html like the below:

```html
<p><strong>foo</strong><p>
<p><strong><br></strong></p>
```

```html
<p><strong>foo</strong><p>
```

This is because `<strong><br></strong>` is considered as empty dom.

But this doesn't seem consistent with the following example:

```html
<p>foo<p>
<p><br></p>
```

This becomes:

```html
<p>foo<p>
<p><br></p>
```

(does not change)

In this case `<p><br></p>` works as a line separator and isn't considered empty.

I think in the former example we should keep `<p><strong><br></strong></p>` as well because this dom is created by the editor as an empty paragragh when an user is writing in `<strong>` context.

To achieve the above I changed the handling of `excludeBr` flag of `isEmpty` function and pass it as true only when all parents (including top level one below the root node) don't have any siblings. (I also exlucde the case that top level node itself is `<br />` because this shouldn't be removed according to the test cases.)

This change doesn't break any existing test cases and passes the cases I added (which are similar to the first example.)

Thanks

### DIFF
--- a/src/plugins/xhtml.js
+++ b/src/plugins/xhtml.js
@@ -734,7 +734,7 @@
 
 			while (childrenLength--) {
 				if (!isEmpty(childNodes[childrenLength],
-					!node.previousSibling && !node.nextSibling)) {
+					excludeBr && !node.previousSibling && !node.nextSibling)) {
 					return false;
 				}
 			}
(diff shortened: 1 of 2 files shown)


samclarke:
Thanks for the PR! Merged.

PR #535, opened 2016-09-26
Replace only "CTRL" by "CTRL or CMD" to mac user

webevolutis:
In mac, it's CMD+Z to undo


samclarke:
Looks good! Merged, thanks.

PR #550, opened 2016-12-15
Fixed stacking of link,image,youtube and email

TwinFuture:
Solved an issue where if you highlight a link you have inserted and go to insert another link in its place using the drop down in WYSIWYG mode, this will stop the element stacking. Example current output:

```
<a href="http://1stlink.com">
    <a href="http://2ndlink.com">DESCRIPTION</a>
</a>
```

In addition to this, the previous link you have highlighted to change will display the link details in the drop down allowing you to edit the previous link instead of having to retype everything again. The same goes for modifying email, images and YouTube Videos.

I have used https:// for YouTube as http will be classed as insecure in 2017.

### DIFF
--- a/src/lib/defaultCommands.js
+++ b/src/lib/defaultCommands.js
@@ -475,6 +475,11 @@ define(function (require) {
 						insert: editor._('Insert')
 					}, true);
 
+				// Get highlighted image if any.
+				var	$current = $(editor.currentNode()),
+					$anchor  = $current.is('img') ? $current :
+					$current.children('img').first();
+
 				content.find('.button').click(function (e) {
 					var	val    = content.find('#image').val(),
 						width  = content.find('#width').val(),
@@ -500,6 +505,17 @@ define(function (require) {
 				});
 
 				editor.createDropDown(caller, 'insertimage', content);
+
+				// If an image is highlighted insert details in dropdown.
+				if ($anchor.attr('width')) {
+					content.find('#width').val($anchor.attr('width'));
+				}
+				if ($anchor.attr('height')) {
+					content.find('#height').val($anchor.attr('height'));
+				}
+				if ($anchor.attr('src')) {
+					content.find('#image').val($anchor.attr('src'));
+				}
 			},
 			tooltip: 'Insert an image'
 		},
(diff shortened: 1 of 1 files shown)


src/lib/defaultCommands.js
```diff
@@ -596,6 +652,18 @@ define(function (require) {
 				});
 
 				editor.createDropDown(caller, 'insertlink', content);
+
+				// Check it is an email address.
+				if ($anchor.attr('href') &&
+					$anchor.attr('href').substr(0, 7) !== 'mailto:') {
+				// Get details of highlighted and put in drop down.
+					if ($anchor.attr('href')) {
```
samclarke: This looks like it's part of the parent condition, can it ever be false?

src/lib/defaultCommands.js
```diff
@@ -500,6 +505,17 @@ define(function (require) {
 				});
 
 				editor.createDropDown(caller, 'insertimage', content);
+
+				// If an image is highlighted insert details in dropdown.
+				if ($anchor.attr('width')) {
+					content.find('#width').val($anchor.attr('width'));
```
samclarke: As the default is just an empty string, could replace the if with `content.find('#width').val($anchor.attr('width') || '');`.

src/lib/defaultCommands.js
```diff
@@ -574,9 +622,17 @@ define(function (require) {
...
 
-							editor.wysiwygEditorInsertHtml(
-								'<a href="' + url + '">' + text + '</a>'
-							);
+							// Replace existing Link, instead of stacking.
+							if ($anchor.length) {
+								if (text) {
+									$anchor.attr('href', text);
+								}
+								$anchor.attr('href', url);
```
samclarke: Could replace this with just `$anchor.attr('href', text || url);`. Maybe not quite as nice but it is a bit smaller.

src/lib/defaultCommands.js
```diff
@@ -596,6 +652,18 @@ define(function (require) {
 				});
 
 				editor.createDropDown(caller, 'insertlink', content);
+
+				// Check it is an email address.
+				if ($anchor.attr('href') &&
```
samclarke: Could you assign `$anchor.attr('href')` to a variable as that should help with minification. Variables can be minified more where as multiple `.attr('href')` can't be and should be less work for the browser.

src/lib/defaultCommands.js
TwinFuture: Oops, that shouldn't be there as its previously checked my bad.. Lol. Its also present on the mail one.. I think the reason i did this is because I had to check one isn;t an email so it didn't grab the mailto string or remove that some chars from the link if link was selected.

Also we should replace 
`$anchor.attr('href').substr(0, 7) !== 'mailto:')`
With
`$anchor.attr('href').indexOf('mailto:') < 0;`
Its shorter and faster (Slightly/maybe lol)! sorry for not putting it in the first place.

Thanks for the corrections!

PR #551, opened 2016-12-15
#fixes 545 Replaced <p> with div for body content.

TwinFuture:
As stated in issue replaced the tags that get inserted into the document on load and when creating a new line. This solves the issue of quotes splitting and removing <p> that would get replaced with <div> so just use div instead as default for content lines.

### DIFF
--- a/src/lib/SCEditor.js
+++ b/src/lib/SCEditor.js
@@ -1996,7 +1996,7 @@
 		 */
 		base.setWysiwygEditorValue = function (value) {
 			if (!value) {
-				value = '<p>' + (IE_VER ? '' : '<br />') + '</p>';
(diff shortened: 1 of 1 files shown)


samclarke:
Thanks for the PR! This fixes the issue for BBCode but not for XHTML output which could have `<p>` tags pasted or added in the source. Also it's currently possible to insert HTML into the BBCode editor or tags that convert to `<p>` which this wouldn't fix.

As `<div>` tags seem to work slightly better it might be worth converting all `<p>` tags into divs with a special class in `setWysiwygEditorValue`. The BBCode plugin works fine with divs and the XHTML plugin can convert them to paragraphs. Will need some testing to verify that works OK but should fix all the issues if it does.

PR #581, opened 2017-11-24
Numeric and URL fields should be LTR, even in RTL mode

abetis:
RTL editor is usually be on RTL page, where <html dir="rtl"> tag will specify right-to-left direction, meaning EVERYTHING on the page will be RTL.
There are some cases that have to be forced to LTR to show in correct direction.


samclarke:
Looks good! Merged. Thanks for the PR.

PR #585, opened 2017-11-25
Youtube tag enhancements

abetis:
The PR includes:
- Change in APIs to allow bbcode formatter access the global options and templates
- youtubeParameters option added to allow people use different video sizes and other formatting options in the generated HTML code
- The formatter now uses the same youtube HTML template as the editor, and the youtubeParameters option
- YouTube start playing time is now supported on both the editor and the formatter
- Added processing of "time_continue" URL parameter as a start time, its added by youtube player when switching from an embedded player to youtube's page big player.
- Unit tests updated according to the API changes
- Added unit tests to check YouTube start time

The API change now allows storing all the HTML used in the parser in the templates, that could be a good idea to have everything in central place since there are some duplicates in the editor and the parser...

### DIFF
--- a/src/formats/xhtml.js
+++ b/src/formats/xhtml.js
@@ -185,13 +185,11 @@
 					editor,
 					caller,
 					function (id, time) {
-						editor.insertText(
-							'<iframe width="560" height="315" ' +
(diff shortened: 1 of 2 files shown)


abetis:
Replaced by #687

PR #587, opened 2017-11-27
List enhancements

abetis:
The change include:
- Drop-down menus to select bullets and numbered lists types
- Both lists types use the HTML5 styling version, although OL can still use the type attribute, the styling version allows more options (I personally need the addition numbering option that I'll add on my local-adaptations branch).
- Unit tests updated accordingly
- Added UTs for different listing types


abetis:
Replaced by #690

PR #591, opened 2017-11-30
List configurable options

abetis:
This change allows configuration of the bullet and ordered list options shown in the drop-down list.
(Depended on other PRs I've sent before).

Allows users to use the proper list types they need, as specified here:
https://www.w3schools.com/cssref/pr_list-style-type.asp


abetis:
Replaced by #690

PR #594, opened 2017-12-01
Inline code implementation

abetis:
This PR include:
- Inline code tags
- [c][/c] tags are used
- 
- The implementation is configurable by allowInlineCode option, disabled by default to keep current behaviour unchanged.

- Some text was selected
- The selected text doesn't include the <p> (full line was selected, or more than one line)
- Not the whole line was selected (happens when selecting a line from the beginning to the end, but the end of the line is not selected, so it's not catched by previous point)

Open issue:
Don't know how to prevent styling inside the inline code segment, something that is done in the [code][/code] segment already.
The styling is stripped after parsing, but the [code] prevents the user buttons while in the segment...

### DIFF
--- a/src/lib/defaultCommands.js
+++ b/src/lib/defaultCommands.js
@@ -479,10 +479,6 @@ var defaultCmds = {
 			var selected = editor.getRangeHelper().selectedHtml();
 			var wholeLine = false;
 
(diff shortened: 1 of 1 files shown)


abetis:
Replaced by #672

PR #596, opened 2017-12-02
Facebook video

abetis:
This PR includes Facebook video support.
- Additional FB video button
- UTs added

Thoughts:
If other video providers are added (Vimeo?), it will probably be wiser to have one button for video addition, that will open a drop down list where user could select the video provider...


abetis:
Replaced by #694

PR #610, opened 2017-12-17
Heading tags

abetis:
This PR include:
- Heading tags logic (H1-H6)
- Drop down list to select the heading type
- Configuration to specify what types should be shown in the drop-down list

An issue I don't know how to solve:
When standing on the heading text, the bold toolbar button is selected.
The browser thinks the highlighting matches the bold text (it is shown as bold, but is wrapped with H tags).

### DIFF
--- a/src/themes/inc/defaultbase.less
+++ b/src/themes/inc/defaultbase.less
@@ -284,7 +284,12 @@ div.sceditor-dropdown div {
 			text-decoration: none;
  		}
 
- 		.sceditor-heading-option h1, h2, h3, h4, h5, h6 {
+ 		.sceditor-heading-option h1,
+ 		.sceditor-heading-option h2,
+ 		.sceditor-heading-option h3,
+ 		.sceditor-heading-option h4,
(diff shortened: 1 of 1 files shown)


bassim:
Hi there! Is there an ETA for the merge of this PR?

brunoais:
I can't give you one because I'm leaving such kinds of decisions to @samclarke

abetis:
Wasn't aware this plugin exist... Will check it later.
So the general movement is to leave a lean engine and plugins that will handle the formatting?

abetis:
@samclarke 
That plugin doesn't work for bbcode, what is the reason for that?
```
// Don't enable if the BBCode plugin is enabled.
if (opts.format && opts.format === 'bbcode') {
	return;
}
```

Will appreciate if you can point me to a documentation explaining the plugin basics, what functions should it implement for example?

Also, I don't like the toolbar button position override, why it's actually needed?
The "format" button could be added to the `toolbar` option as any other button...

abetis:
Replaced by #685

PR #661, opened 2018-03-14
Always uses plain text to generate YouTube embed in AutoYouTube plugin

### LINKED ISSUE #660: AutoYouTube plugin wraps iframe in unwanted font and a elements
After enabling the AutoYouTube plugin, pasting a YouTube url into the editor does indeed create the embed, but also wraps it in several layers of `<font>` elements and an `<a>` element, like so:
```
<font size="3"><font face="Helvetica"><font size="3"><font face="Helvetica"><a href="https://www.youtube.com/watch?v=UnPI4xKvf0c"><iframe width="560" height="315" frameborder="0" src="https://www.youtube.com/embed/UnPI4xKvf0c?wmode=opaque" data-youtube-id="UnPI4xKvf0c" allowfullscreen=""></iframe></a></font></font></font></font>
```

It would be better if the generated HTML were simply the `<iframe>` element, like so:
```
<iframe width="560" height="315" frameborder="0" src="https://www.youtube.com/embed/UnPI4xKvf0c?wmode=opaque" data-youtube-id="UnPI4xKvf0c" allowfullscreen=""></iframe>
```

Sesquipedalian:
Fixes #660

### DIFF
--- a/src/plugins/autoyoutube.js
+++ b/src/plugins/autoyoutube.js
@@ -76,14 +76,10 @@
 				return;
 			}
 
-			if (data.html || data.text) {
+			if (data.text) {
 				var html = document.createElement('div');
 
-				if (data.html) {
-					html.innerHTML = data.html;
-				} else {
-					html.textContent = data.text;
-				}
+				html.textContent = data.text;
 
 				convertYoutubeLinks(html);
 


brunoais:
Your PR is not solving the problem.
If you try pasting two youtube videos with a paragraph between them, the results won't be satisfactory.
Both youtube videos will be places side-by side (when one is not wrapped to the next like).
Same happens if you are pasting paragraphs of text and there's youtube videos in between.

Instead of what you are doing, to fix this, what makes sense is to use heuristics (even if simple ones) to determine what to delete.
Maybe just deleting all markup HTML that only goes around the youtube video is enough.

Sesquipedalian:
@brunoais Did you test this before you commented? The results are shown below:
screen shot 2018-03-14 at 2 30 58 am

Sesquipedalian:
If there are ways to produce unsatisfactory results using this PR, please post instructions for how to reproduce them.

Sesquipedalian:
Pasting `This is a video https://www.youtube.com/watch?v=G2y8Sx4B2Sk Isn't it fun?` produces satisfactory results.
screen shot 2018-03-14 at 2 42 26 am

---

Pasting a string of text that contains line breaks never produces satisfactory results, regardless of whether a YouTube link is embedded into it or not. This problem is not due to any of the changes introduced in this PR. The same behaviour occurs without this PR's changes.
```
Paragraph 1
https://www.youtube.com/watch?v=G2y8Sx4B2Sk
Paragraph 3
```
screen shot 2018-03-14 at 2 56 26 am

```
Paragraph 1
Paragraph 2
Paragraph 3
```
screen shot 2018-03-14 at 3 06 14 am

---

Finally, if one pastes a string containing two YouTube links, only the first will be transformed into an iframe, but this is true of the existing code and is not due to any of the changes in this PR.

---

Based on these tests, I can see no problematic results arising from this PR's proposed changes. The only continuing problems I have found are ones that pre-existed this PR.

brunoais:
To tell the truth, I didn't test your code in this PR.
In that case, I'll let @samclarke decide

Sesquipedalian:
Closing this since, for reasons I am not sure of, the error no longer occurs for me as of version 2.1.3.

Sesquipedalian:
Oh, wait. The error does still occur. But this PR no longer seems to fix it properly.

Issue #662, opened 2018-03-15
Add min-width to inserted table (usability)


q2apro:
With default theme (and I guess all others), inserting a 2x2 table gives:

![image](https://user-images.githubusercontent.com/6764548/37462761-b667fba0-285b-11e8-97ab-35d4a5d56fa5.png)

Adding: 

     table td {
       min-width:30px;
     }

gives:

![image](https://user-images.githubusercontent.com/6764548/37462786-c745385c-285b-11e8-921f-280670139155.png)

tata81:
Hi @q2apro , I am new to open source contribution, would like to work and contribute to this issue. Thanks

brunoais:
@q2apro I'm more into using `15px` instead of `30px` What do you think?

PR #663, opened 2018-03-23
Added min-width to inserted table.

tata81:
re #662 Given width width of 30px to the table in default.css file.


brunoais:
set it to `5px` and I'll merge

PR #666, opened 2018-04-01
Checks wysiwygWindow.getSelection() !== null before getting rangeCount

### LINKED ISSUE #665: Error if wysiwygWindow.getSelection() is null
Under certain circumstances (see https://github.com/SimpleMachines/SMF2.1/issues/4685), an error may be encountered in checkSelectionChanged if the check() function tries to perform `wysiwygWindow.getSelection().rangeCount` when `wysiwygWindow.getSelection()` is null. 

For some reason, Firefox reports the error but WebKit browsers do not.

Sesquipedalian:
Fixes #665


samclarke:
Looks good! Thanks for the PR 👍 Merged.

PR #670, opened 2018-04-05
Code block should always be in LTR direction.

abetis:
Code segment should always be LTR, otherwise its incorrectly shown in RTL display mode.


samclarke:
Thanks for the PR! Merged.

Issue #648, opened 2018-02-13
Multiple whitespaces not preserved in <code> block


q2apro:
It is essential for my users to write code and indent it by whitespaces, see example of code at https://www.stacklounge.de/4/

However, with Sceditor 2.1.2 the whitespaces are not preserved, they are instead removed.

Should this be considered a bug?

abetis:
I have the same issue.
Couldn't figure out yet when the spaces are shown and when are not.
Sometimes I just re-edit several times, sometimes switch to code mode and the spaces do appear properly...
Probably something related to trimming somewhere in the code...

q2apro:
Or should we insert a `<pre><code>` instead of only `<code>`? 

This way the spaces would survive. But my test was not successful. I am wondering if SCEditor removes those \n line breaks within the sceditor textarea on submit.

samclarke:
The white space hasn't been removed but it is being collapsed by the browser.

If you add:

```css
code {
    white-space: pre;
}
```

to the linked page it will show the spaces. You're right though, for HTML output it would probably make more sense for the editor to output `<pre><code>` instead of just `<code>`.

q2apro:
Thanks for this tip, it works now. Even without <pre> tag. 

However, each line gets a `<br>` in the beginning, which now leads two two lines, even though it is only one line set. 

**Input:** 

![image](https://user-images.githubusercontent.com/6764548/36340753-a7533ec4-13eb-11e8-8516-1ac530eaf35a.png)

**Parsed to:** 

`<p>
	<code>for (var i = 0 / 2; i &lt; classes.length; i++) <br />{<br />  if(checkCondition(classes[i]) === undefined) <br />  {<br />    console.log('undefined');<br />  }<br />}<br /></code>
</p>`

which then looks like this on the website: 

![image](https://user-images.githubusercontent.com/6764548/36340756-c37b0f3c-13eb-11e8-8ba2-32b2b17d8be3.png)

https://www.stacklounge.de/1045/stacklounge-editor-vorschlage-verbesserung-optimierung

**My first idea to fix:** 

```
code br {
    display: none;
}
```

abetis:
Double newlines is a big problem for me as well.
I reported it few months ago, hope Sam can address it soon.

q2apro:
**Suggestion / Solution:**

1. When a user hits the code button, then always insert `<pre><code>` (like stackoverflow is doing it)
2. Never add `&nbsp;` or `<br>` when within pre-code-element.

[...]

PR #672, opened 2018-04-05
Inline code implementation

abetis:
This PR include:
    Inline code tags
    [c][/c] tags are used
    The implementation is configurable by allowInlineCode option, disabled by default to keep current behaviour unchanged.

    Some text was selected
    The selected text doesn't include the

    (full line was selected, or more than one line)
    Not the whole line was selected (happens when selecting a line from the beginning to the end, but the end of the line is not selected, so it's not catched by previous point)

Open issue:
Don't know how to prevent styling inside the inline code segment, something that is done in the [code][/code] segment already.
The styling is stripped after parsing, but the [code] prevents the user buttons while in the segment...

### DIFF
--- a/src/formats/bbcode.js
+++ b/src/formats/bbcode.js
@@ -168,6 +168,9 @@
 		code: {
 			txtExec: ['[code]', '[/code]']
 		},
+		c: {
+			txtExec: ['[c]', '[/c]']
+		},
 		image: {
 			txtExec: function (caller, selected) {
 				var	editor  = this;
@@ -715,6 +718,44 @@
 		},
 		// END_COMMAND
 
+		// START_COMMAND: c (inline code)
+		c: {
+			tags: {
+				c: null
+			},
+			isInline: true,
+			allowedChildren: ['#'],
+			format: '[c]{0}[/c]',
+			html: '<span class="inline-code" data-inline-code="1">{0}</span>'
+		},
+		// END_COMMAND
+
+		// START_COMMAND: c (inline code)
+		span: {
+			tags: {
+				span: {
+					'data-inline-code': null
+				}
+			},
+			isInline: true,
+			allowedChildren: ['#'],
+			format: function (element, content) {
+				if (this.opts.allowInlineCode) {
+					return '[c]' + content + '[/c]';
+				} else {
+					return content;
+				}
+			},
+			html: function (token, attrs, content) {
+				if (this.opts.allowInlineCode) {
+					return '<span class="inline-code" data-inline-code="1">' +
+						content + '</span>';
+				} else {
+					return content;
+				}
+			}
+		},
+		// END_COMMAND
 
 		// START_COMMAND: Left
 		left: {
@@ -2256,7 +2297,8 @@
 			ol: ['li', 'ol', 'ul'],
 			table: ['tr'],
 			tr: ['td', 'th'],
-			code: ['br', 'p', 'div']
+			code: ['br', 'p', 'div'],
+			span: ['br']
 		};
 
 		/**
--- a/src/lib/defaultCommands.js
+++ b/src/lib/defaultCommands.js
@@ -442,10 +442,33 @@ var defaultCmds = {
 	// START_COMMAND: Code
 	code: {
 		exec: function () {
-			this.wysiwygEditorInsertHtml(
-				'<code>',
-				(IE_BR_FIX ? '' : '<br />') + '</code>'
-			);
+			var editor = this;
+			var range = editor.getRangeHelper().selectedRange();
+			var selected = editor.getRangeHelper().selectedHtml();
+			var wholeLine = false;
+
+			// Check if the whole line was selected
+			// It could be the whole text of the text type node (3)
+			// or innerHTML of an element node (1) if there were some markups
+			// in the text, so its containted in a paragraph <p>...</p>
+			if ((range.commonAncestorContainer.nodeType === 3 &&
+				range.commonAncestorContainer.wholeText === selected) ||
+				(range.commonAncestorContainer.nodeType === 1 &&
+				range.commonAncestorContainer.innerHTML === selected)) {
+				wholeLine = true;
+			}
+
+			if (editor.opts.allowInlineCode && !wholeLine && selected &&
+				selected.length > 0 && !selected.includes('<p>')) {
+				this.wysiwygEditorInsertHtml(
+					'<span class="inline-code" data-inline-code="1">', '</span>'
+				);
+			} else {
+				this.wysiwygEditorInsertHtml(
+					'<code>',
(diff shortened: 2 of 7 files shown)


src/formats/bbcode.js
```diff
@@ -715,6 +718,44 @@
...
+		span: {
+			tags: {
+				span: {
+					'data-inline-code': null
+				}
+			},
+			isInline: true,
+			allowedChildren: ['#'],
+			format: function (element, content) {
+				if (this.opts.allowInlineCode) {
```
brunoais: Remove this condition.
**because:** If it is meant not to be allowed, then it is the website developer's responsibility to remove the BBCode prior to attaching to the textarea.

src/formats/bbcode.js
```diff
@@ -715,6 +718,44 @@
...
+					return '[c]' + content + '[/c]';
+				} else {
+					return content;
+				}
+			},
+			html: function (token, attrs, content) {
+				if (this.opts.allowInlineCode) {
```
brunoais: Idem

src/lib/defaultCommands.js
```diff
@@ -442,10 +442,33 @@ var defaultCmds = {
...
+				selected.length > 0 && !selected.includes('<p>')) {
+				this.wysiwygEditorInsertHtml(
+					'<span class="inline-code" data-inline-code="1">', '</span>'
+				);
+			} else {
+				this.wysiwygEditorInsertHtml(
+					'<code>',
+					(IE_BR_FIX ? '' : '<br />') + '</code>'
+				);
+			}
```
brunoais: This new doesn't work and it is not as easy as it looks. Better leave that to two buttons.
A user may want to transform a single line of code into a code block.
E.g. How do you manage a press of `Enter` while inside the inline-code block when the user had used the code block? I can't find such code.

src/formats/bbcode.js
abetis: I'll happily remove the configuration option since I always need that feature enabled.
On the other side, the website developer won't have an option to disable that feature.

Its will be either all the code button functionality or nothing.

brunoais: the website developer can delete the BBCode himself. It is in the API documentation.

src/lib/defaultCommands.js
brunoais: I meant that the concept doesn't work, not the code itself. (the reason is in the prev comment)

src/formats/bbcode.js
abetis: Remove BBCode?! By editing the JavaScript of the SCEditor?!
I can bet 90% of the SCEditor users are probably those who can only follow the instruction as copy&paste and the advanced of them can change the configuration options.
Removing inline code, which is twisted in the code implementation will not be easy.
The options allow to exclude some buttons, so its either no code button at all or have the button for code block and inline code alltogether, which is fine for me since I always need both of the options, and I believe users who add code snippets to their comments will say the same.

I'll remove the option in my next push.

src/lib/defaultCommands.js
abetis: The code marks the selected text as code. I can't see any inconsistence.
Play with the code button here on GitHub or phpBB forum and you will see exactly the same behavior.
Unfortunately I couldn't find CKEditor demo with code button, so I don't know how they do that.

The proposed behavior was very intuitive for me, while editing about 100 pages with code snippets all over. And consistent with GitHub and phpBB behavior as I wrote...

Anyway, don't integrate the PR yet, I want eliminate the use of <span> and move to <code> for inline and <pre><code> for code blocks, it will address issue #648 as well and the implementation will probably be better.
I'll send it as a separate PR.

src/formats/bbcode.js
brunoais: > Remove BBCode?! By editing the JavaScript of the SCEditor?!
No. By calling `sceditor.formats.bbcode.remove('c')`

src/lib/defaultCommands.js
brunoais: the `[c]` and `[code]` bbcodes in phpBB central forum are handled as different BBCodes and they are always clearly different.
Here, you are joining the inline code with the block code as a single command. That is my problem with this. If you make a new command for the inline code, then I have no issues with this.

abetis: Sorry, but I think we are talking about something different.

The code introduce `[c]` bbcode in addition to `[code]` that is already implemented.
The code decides whether to use `[c]` or `[code]` based on the selected text:
full line selected - use `[code]`
text in a line - use `[c]`

Currently the `[code]` block is marked with `<code>` HTML tag, while `[c]` tag is marked with `<span class="inline-code">` in non-source view.

`[c]` and `[code]` are absolutely different codes, so I don't understand your rejections...

brunoais: I'll try like this:
How do you show a user the difference between both outcomes?
When a user presses `Enter`, the outcome isn't the same depending on both possible situations.
enter_on_inline_code
Because, if the new line is already there, the style is removed and reapplied....
if_unstyled_and_restyled
The outcome doesn't fit with the previous state.

samclarke:
I think merging this should wait until after looking into #676 otherwise this will need an option to enabled it to be completely backwards compatible (I doubt any one is using `[c]` for anything else but it's possible). If dialects are implemented then this can be included in the right dialects and be completely backwards compatible.

brunoais:
@samclarke there is no `[c]` BBCode in phpBB by default.

src/lib/defaultCommands.js
abetis: I will check why your last example didn't choose the old code block style, but I'm ok with the rest of the behaviour...

abetis:
Guys, I want to continue with all the changes I need for my implementation, but I don't want to create and maintain a total separate fork. I think you and many other users will want to use those features.

@samclarke 
Please advise a way to propose you the changes, so I won't have to implement it 3'rd time...
I understand that plugins are a better way, so 2 answers are needed in that direction:
1: Can a plugin override current tag implementation? Can I add a plugin that overrides current `[code]` tag implementation?
2: In order not to duplicate the implementations of the same thing, how can I call the default `[code]` implementation from my plugin for non-inline code blocks?

Thanks.

abetis:
@brunoais 
I thought it was a default, but apparently it was just added as a custom tag.
It exist on phpbb.com as well, just try to preview the following in their forum:
`bla bla [c]code[/c] bla bla`
They have a "c" button in the toolbar as well.

brunoais:
> Can a plugin override current tag implementation?

[Yes BBcodes](https://www.sceditor.com/documentation/custom-bbcodes/) and [Yes commands](https://www.sceditor.com/documentation/custom-commands/)
Dunno the 2nd one.

> I thought it was a default, but apparently it was just added as a custom tag. [...] They have a "c" button in the toolbar as well.

Yep.

Issue #673, opened 2018-04-05
Unexpected sceditor-nlf paragrapth with break line after unordered and ordered list


pnowy:
Hi,

we are using sceditor and we noticed unexpected behavior.

When we create simple ordered/unordered list, for example:

- one
- two
- three

and after that list we start new paragraph (by creating next list item and uncheck the list button) the sceditor create unexpected break line within paragraph with weird class. Source HTML:

```
<div>
	<ul>
		<li>
			one 
		</li>
		<li>
			two 
		</li>
		<li>
			three 
		</li>
	</ul><br /> 
</div>
<p class="sceditor-nlf">
	<br />
</p>
```

![change_proportion___dietmap_admin_panel___env__local___ver__1_0](https://user-images.githubusercontent.com/3254609/38382256-a7d9ba2e-3909-11e8-90db-9ebb2b394528.png)

Could you tell what is the purpose of the additional HTML code generated and is there an option to disable this behavior (unfortunately we are working a lot with lists and we would like to remove this additional HTML).

Thanks

brunoais:
That paragraph is there so that the person editing the content can start typing after the list if he chooses to.
It shouldn't appear in the resulting xhtml or BBCode. If it does, then it is a bug.

brunoais:
Which version of SCE are you using?

pnowy:
Version 2.1.0

brunoais:
It's a bug. Either the `xhtml` plugin or SCE's internals will have to be changed accordingly.
Will be handled when able.

samclarke:
Thanks for reporting! This is now fixed in the master branch by PR #675

PR #675, opened 2018-04-06
Remove empty "sceditor-nlf" paragraph in XHTML format

abetis:
Updating XHTML format with the code from bbcode format to remove empty "sceditor-nlf" paragraphs.
Addresses #673

### DIFF
--- a/src/formats/xhtml.js
+++ b/src/formats/xhtml.js
@@ -736,6 +736,12 @@
 							/br/i.test(firstChild.nodeName))) {
 							// Mark as empty,it will be removed by the next code
 							empty = true;
+						} else {
+							node.classList.remove('sceditor-nlf');
+
+							if (!node.className) {
+								removeAttr(node, 'class');
+							}
 						}
 					}
 				}
--- a/src/formats/xhtml.js
+++ b/src/formats/xhtml.js
@@ -12,6 +12,12 @@
 (function (sceditor) {
 	'use strict';
 
+	var IE_VER = sceditor.ie;
+
+	// In IE < 11 a BR at the end of a block level element
+	// causes a double line break.
+	var IE_BR_FIX = IE_VER && IE_VER < 11;
+
 	var dom = sceditor.dom;
 	var utils = sceditor.utils;
 
@@ -707,6 +713,7 @@
 						isTopLevel && noSiblings && tagName !== 'br'),
 					document        = node.ownerDocument,
 					allowedTags     = xhtmlFormat.allowedTags,
+					firstChild   	= node.firstChild,
 					disallowedTags  = xhtmlFormat.disallowedTags;
 
 				// 3 = text node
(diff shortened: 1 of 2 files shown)


src/formats/xhtml.js
```diff
@@ -720,6 +727,19 @@
...
+				if (nodeType === 1) {
+					// skip empty nlf elements (new lines automatically
+					// added after block level elements like quotes)
+					if (is(node, '.sceditor-nlf')) {
+						if (!firstChild || (!IE_BR_FIX &&
+							node.childNodes.length === 1 &&
+							/br/i.test(firstChild.nodeName))) {
+							// Mark as empty,it will be removed by the next code
+							empty = true;
+						}
```
samclarke: Thinking about it, if the `sceditor-nlf` node isn't empty then it should probably remove the class. So something like:

```js
else {
	node.classList.remove('sceditor-nlf');

	if (!node.className) {
		removeAttr(node, 'class');
	}
}
```

abetis:
@samclarke 
The UT bumps into another problem:
The `stripWhiteSpace` in test/unit/utils.js strips all the newline, tabs and spaces as well, producing the following:
>> Message: None empty with extra class
>> Actual: "<divclass=\"test\">test</div>"
>> Expected: "<div class=\"test\">test</div>"

Removing the `\s` from RegEx will fix that problem, but will fail many other tests where attributes have spaces, such as `font-size: 24px;`, where the expected results are without the space after ":".

It makes sense to check that the attributes don't have the space, but you can't strip the space from the whole element string as its done now...

What do you want to me to do with that?
Remove the `\s` from the RegEx and fix all other expected results to include the space after ":"?

abetis:
If you really want to remove all the spaces in the attributes, we should probably do it in the `handleElement` function and replace that line:
`attrValue = attr.value;
`
with something like:
`attrValue = attr.value.replace(/[\s]/g, '');
`
but there are still few tests that fail since they have spaces in the expected results.
@samclarke , your decision for the direction. I'm ok with the spaces in attributes,
such as: `font-size: 24px;`, but removing the space in the whole HTML element string is definitely wrong.

There are some UTs that have the expected results without the space, which is absolutely wrong:
>> plugins/xhtml - Should not empty tags with a size
>> Message: Empty tag with size
>> Actual: "<p><i style=\"display:inline-block;width:1px;height:1px;\"></i></p>"
>> Expected: "<p><**istyle**=\"display:inline-block;width:1px;height:1px;\"></i></p>"

samclarke:
`filterStripWhiteSpace` strips all spaces for tests like the one above that just verifies that the attribute is removed. If you really want to verify the white space then you can use the `filterHtml` method instead and add the extra white space but for that test it doesn't matter.

abetis:
Ohh... I've missed that the strip is in the UT code and not in the application.
Updated the files.

samclarke:
Merged. Thanks!

PR #684, opened 2018-04-09
Inline code plugin implementation

4project-co-il:
This PR include:
- Implementation of the inline code tags
- [c][/c] tags are used
- Plugin named "code" is used to include that implementation
- The plugin overrides current implementation of the code tag
- The inline code is marked with `<code style="inline">` in the WYSIWYG mode

- Some text was selected
- The selected text doesn't include the <p> (full line was selected, or more than one line)
- Not the whole line was selected (happens when selecting a line from the beginning to the end, but the end of the line is not selected, so it's not catched by previous point)

No UTs since I couldn't find an example how to initiate plugin before running the test.

### DIFF
--- a/src/plugins/inline-code.js
+++ b/src/plugins/inline-code.js
@@ -21,7 +21,7 @@
 	// causes a line break. In all other browsers it's collapsed.
 	var IE_BR_FIX = IE_VER && IE_VER < 11;
 
-	sceditor.plugins.code = function () {
+	sceditor.plugins['inline-code'] = function () {
 		var base = this;
 
 		/**
--- a/tests/manual/debug/index.html
+++ b/tests/manual/debug/index.html
@@ -23,7 +23,7 @@
 		<script src="../../../src/plugins/plaintext.js"></script>
 		<script src="../../../src/plugins/autosave.js"></script>
 		<script src="../../../src/plugins/dragdrop.js"></script>
-		<script src="../../../src/plugins/code.js"></script>
+		<script src="../../../src/plugins/inline-code.js"></script>
 		<script src="../../../src/icons/monocons.js"></script>
 		<script src="../../../src/icons/material.js"></script>
 		<script src="../../../src/icons/iconmoon.js"></script>


src/lib/dom.js
```diff
@@ -741,7 +741,7 @@ export function isInline(elm, includeCodeAsBlock) {
 
 	tagName = elm.tagName.toLowerCase();
 
-	if (tagName === 'code') {
+	if (tagName === 'code' && !is(elm, '.inline')) {
```
brunoais: I have some doubts about this.... But I do understand it is a hard situation you were placed in due to the previous code....
I'll leave such decision to @samclarke

src/plugins/code.js
```diff
@@ -0,0 +1,155 @@
...
+		base.init = function () {
+			var	opts  = this.opts;
+
+			// Enable for BBCode only
+			if (opts.format && opts.format !== 'bbcode') {
+				return;
+			}
+
+			// The plugin will override current code implementation
+			this.commands.code = {
```
brunoais: Use the public API instead.

src/lib/dom.js
abetis: I don't like that by myself, as well as the hardcoded enumeration I have to use:
>			if ((range.commonAncestorContainer.nodeType === 3 &&
>				range.commonAncestorContainer.wholeText === selected) ||
>				(range.commonAncestorContainer.nodeType === 1 &&
>				range.commonAncestorContainer.innerHTML === selected)) {
>				wholeLine = true;
>			}
But that's the current infrastructure limitation and I cant/want change it in order to implement a relatively simple feature.
That's the reason I report the issues in hope that it will be changed one day.

I agree that using plugins is better in that case since it has minimal affect on the core code, but the environment is not ready yet for plugins development as well, so I'll have to hit my head on a wall...

src/plugins/code.js
abetis: updated

brunoais:
Now it's up to @samclarke to evaluate if all the rest looks good to him.

abetis:
It's more complicated than I thought since there are more places where code tag is handled in a hardcoded fashion, so the tags are stripped when switching to source view and back.
Will push the updated code soon.

abetis:
Apparently the problem was wrong `isInline` value, but I've added also some logic to handle inline code in source edit mode.
I've also fixed the plugin initialization after pulling my hear out since sometimes the plugin worked and sometimes not.
After some debugging I saw that the plugin never worked when the page first loads, but did work when I press the "Create Editor" button on the manual debug page...

abetis:
Only filename change

abetis:
> If there's bugfixes for SCE, do them in a separate PR, pls
#686

PR #685, opened 2018-04-10
Adding bbcode support to format plugin

abetis:


src/plugins/format.js
```diff
@@ -62,12 +63,93 @@
...
+
+				sceditor.formats.bbcode.set('h6', {
+					tags: {
+						h6: null
+					},
+					isInline: false,
+					allowedChildren: ['#'],
+					format: '[h6]{0}[/h6]',
+					html: '<h6>{0}</h6>'
+				});
```
brunoais: May you do the headers in a for loop?

abetis:
I wanted to do it, but I don't know how a tagname variable can be converted to property name:
```
+					tags: {
+						h1: null
+					},

```

brunoais:
var obj = {
				tags: {
				},
				isInline: false,
				allowedChildren: ['#'],
				format: '[h' + num + ']{0}[/h' + num + ']',
				html: '<h' + num + '>{0}</h' + num + '>'
			}
	obj['tags']['h' + num] = null
	return obj

^inside a function (not tested)

abetis:
Updated. thanks.
That's one of the reasons I don't like JS... Too many ways to get to the same thing.

src/plugins/format.js
```diff
@@ -62,12 +63,27 @@
...
-			if (!this.commands.format) {
-				this.commands.format = {
-					exec: formatCmd,
-					txtExec: formatCmd,
-					tooltip: 'Format Paragraph'
-				};
+			sceditor.command.set('format', {
+				exec: formatCmd,
+				txtExec: formatCmd,
+				tooltip: 'Format Paragraph'
```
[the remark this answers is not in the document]
brunoais: It would have been:
`tooltip: instance._('Format Paragraph')`
but I had failed to notice that SCE already calls it automatically at the time the buttons are generated.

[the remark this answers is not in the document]
abetis: ok. I've copied that as is from defaultCommands.js.
So besides that everything is ok?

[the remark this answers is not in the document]
brunoais: Yes. I didn't have the ability to test, unfortunately. My life has been hellish to find any time frame for more than 5 mins at random times.... Which I need to test these stuff before merging.

src/plugins/format.js
```diff
@@ -33,6 +33,19 @@
...
+			h2: 'Heading 2',
+			h3: 'Heading 3',
+			h4: 'Heading 4',
+			h5: 'Heading 5',
+			h6: 'Heading 6',
+			address: 'Address',
+			pre: 'Preformatted Text'
+		};
+
+		var tags;
```
brunoais: Add a small comment explaining why this is here. Ppl may forget to check the latter lines to understand.

abetis: done

brunoais: I meant the `var tags;` code

abetis: Sorry, don't understand what do you mean...
Explain in comments why tags variable is needed?!

brunoais: Somewhat.
It is explaining what the `tags` variable is used for.

abetis: updated

brunoais: > with the list **form** bbcodeTags or

Change to:
with the list **from** bbcodeTags or

abetis: updated

PR #686, opened 2018-04-10
Fix plugins initialization problems

4project-co-il:
Plugins change defaultCommands and formatter handlers on initialization.
The internal commands list should be rebuild affter plugins initialization, as well as the formatter init, which builds cache based on the handlers that were added/changed by the plugins.


live627:
- Closing because this was already addressed in https://github.com/samclarke/SCEditor/pull/822

PR #688, opened 2018-04-11
percentage-fontsize implementation

abetis:
The plugin allows specifying font size as a percent value as used in phpBB system.


src/lib/SCEditor.js
```diff
@@ -2250,6 +2250,18 @@ export default function SCEditor(original, userOptions) {
...
+	/**
+	 * Returns list of tags from the WYSIWYG editor
+	 *
+	 * @param {string} tagName
+	 * @function
+	 * @name execCommand
+	 * @memberOf SCEditor.prototype
+	 */
+	base.wysiwygGetTags = function (tagName) {
+		return wysiwygDocument.getElementsByTagName(tagName);
```
brunoais: This returns a **live** ordered collection of **Elements** by tag name.
Either be explicit about it in its documentation or change the code into something most programmers would expect from it

src/plugins/percentage-fontsize.js
```diff
@@ -0,0 +1,134 @@
...
+
+	sceditor.plugins['percentage-fontsize'] = function () {
+		var base = this;
+
+		/**
+		 * Default sizes
+		 * @type {Object}
+		 * @private
+		 */
+		var sizes = [50, 75, 100, 150, 200];
```
brunoais: Does this behave like a constant? If so, turn the variable name into **CAPS**
Personally, I'd have a default value but allow setting options to change it

src/plugins/percentage-fontsize.js
```diff
@@ -0,0 +1,134 @@
...
+			// Enable for BBCode only
+			if (opts.format && opts.format !== 'bbcode') {
+				return;
+			}
+
+			// The plugin will override current fontsize implementation
+			sceditor.command.set('size', {
+				exec: commandHandler,
+				txtExec: commandHandler,
+				tooltip: 'Font Size'
```
brunoais: Use `_` function to make it translatable or allow it to be translable somehow.

src/plugins/percentage-fontsize.js
```diff
@@ -0,0 +1,134 @@
...
+				},
+				styles: {
+					'font-size': null
+				},
+				format: function (element, content) {
+					var	fontSize = dom.attr(element, 'size');
+					var size;
+
+					if (!fontSize) {
+						fontSize = element.style['font-size'];
```
brunoais: `element.style['font-size']` -> `element.style.fontSize`

src/plugins/percentage-fontsize.js
abetis: Please explain what do you mean here.
`_` for what function?
What does "translatable" means?

src/lib/SCEditor.js
abetis: I'll check other methods, maybe it will be easier to just inject the HTML tags instead of using execCommand

src/plugins/percentage-fontsize.js
brunoais: It's a function named `_`. See: https://www.sceditor.com/api/sceditor/underscore/
"translatable" <=> To be able to have a different text on the node based on the current language

src/plugins/percentage-fontsize.js
abetis: good point. updating.

src/plugins/percentage-fontsize.js
abetis: Sorry, but I don't understand what function do you mean that have to be with underscore or translatable.
I'm basing my code the same way as is appears in the format plugin that Sam wrote and as defaultCommands.js code looks like...

src/plugins/percentage-fontsize.js
abetis: Is there a documentation for plugin configuration fomat?
Couldn't find any documentation for format.js configuration for example...

src/plugins/percentage-fontsize.js
brunoais: Ah, OK. The code sends tooltips through the translation function when displaying:
https://github.com/samclarke/SCEditor/blob/master/src/lib/SCEditor.js#L712
I had forgotten about that.

PR #592, opened 2017-11-30
Alternative lists

abetis:
This PR include:
- Implementation of alternative list bbcode representations 
(phpBB style, in format `[list=type][*]text[*]text[/list]`)
- Unit tests updated to reflect the fact that there will not be a closing [/*] any more
- Added UTs to test alternative listing representation
- Had to comment out the "Closing parent tag from list item" UT since it conflicts with the non-closing [*] concept. Should probably remove that test if PR is accepted.

TODO:
Investigate why there is a double newline after first [/list] in "Nested UL tag" test.

PR #690, opened 2018-04-12
Enjanced lists plugin implementation

abetis:
The plugin implement:
- Drop-down menus to select bullets and numbered lists types
- Allow configuring the options shown in the drop down lists based on the standard:
https://www.w3schools.com/cssref/pr_list-style-type.asp
- Both lists types use the HTML5 styling version, although OL can still use the type attribute, the styling version allows more options (I personally need the additional numbering option that will be used with custom configuration

### DIFF
--- a/src/plugins/list-enhanced.js
+++ b/src/plugins/list-enhanced.js
@@ -1,0 +1,344 @@
+/**
+ * SCEditor Inline-Code Plugin for BBCode format
+ * http://www.sceditor.com/
+ *
+ * Copyright (C) 2011-2013, Sam Clarke (samclarke.com)
+ *
+ * SCEditor is licensed under the MIT license:
+ *	http://www.opensource.org/licenses/mit-license.php
+ *
+ * @fileoverview SCEditor list enhancement plugin
+ * This plugin expands the implementation of the list commands, adding dropdown
+ * options to select between different list types
+ * @author Alex Betis
+ */
+
+(function (sceditor) {
+	'use strict';
+
+	var dom = sceditor.dom;
+
+	/**
+	* Fixes a bug in FF where it sometimes wraps
+	* new lines in their own list item.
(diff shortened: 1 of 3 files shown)


brunoais:
Please do not link to w3schools. Those guys are too money hungry compared to the actual help they provide. Link to MDN instead.

abetis:
I don't care. Both provide free information on the web. w3 looks more simple to understand.

brunoais:
Because you don't care, I updated for you.

src/plugins/list-enhanced.js
```diff
@@ -0,0 +1,344 @@
...
+	'use strict';
+
+	var dom = sceditor.dom;
+
+	/**
+	* Fixes a bug in FF where it sometimes wraps
+	* new lines in their own list item.
+	* See issue #359
+	*/
+	function fixFirefoxListBug(editor) {
```
brunoais: What prevents you from making an algorithm that works on all browsers?

abetis: I follow the original list commands implementation:
```
	// START_COMMAND: Bullet List
	bulletlist: {
		exec: function () {
			fixFirefoxListBug(this);
			this.execCommand('insertunorderedlist');
		},
		tooltip: 'Bullet list'
	},
	// END_COMMAND
	// START_COMMAND: Ordered List
	orderedlist: {
		exec: function () {
			fixFirefoxListBug(this);
			this.execCommand('insertorderedlist');
		},
		tooltip: 'Numbered list'
	},
	// END_COMMAND
```

abetis:
Added alternative lists logic from #592 
Can be configured to use the OL and UL lists format if someone need it.

src/plugins/list-enhanced.js
brunoais: Fair enough

PR #694, opened 2018-04-16
Video plugin

4project-co-il:
Based on #686 and #687 
The plugin removes youtube command and adds a new "video" command that will recognize which video tag should be used based on the provided URL.

Current implementation includes YouTube and Facebook videos processing.
Youtube bbcode formater is used from the based code, which is based on the #687  PR in order to use the configuration parameters. I could move it to the plugin as well if you decide to move the youtube logic out of the base code.

Material and Monocons youtube icon is used as movie icon. Famfamfam new icon was added.
Movie icon is placed after the picture icon, which makes more sense than current youtube button location.

### DIFF
--- a/src/lib/defaultOptions.js
+++ b/src/lib/defaultOptions.js
@@ -364,7 +364,7 @@ export default {
 			*
 			* @type {string}
 			*/
-		facebookParameters: 'width="560" height="315"' +
+		facebookParameters: 'width="560" height="315" ' +
 			'style="border:none;overflow:hidden" scrolling="no" ' +
 			'frameborder="0" allowTransparency="true" allowFullScreen="true"'
 	},


src/lib/defaultOptions.js
```diff
@@ -350,7 +350,24 @@ export default {
...
+		*/
+		youtubeParameters: 'width="560" height="315" frameborder="0" ' +
+			'allowfullscreen',
+
+		/**
+			* Parameters that will be added to Facebook frame tag
+			*
+			* @type {string}
+			*/
+		facebookParameters: 'width="560" height="315"' +
```
skmedix: Missing space at the end of the string

abetis: thanks

abetis:
Yes. The first line of the PR description:
Based on #686 and #687

brunoais:
OK, got it. Wrong interpretation

src/plugins/video.js
```diff
@@ -0,0 +1,204 @@
...
+
+			var div;
+			var label;
+			var input;
+			var button;
+
+			div = document.createElement('div');
+			label = document.createElement('label');
+			label.setAttribute('for', 'link');
+			label.textContent = editor._('Video URL:');
```
brunoais: Nice image

PR #701, opened 2018-07-15
Fix going back in the browser causes content to be lost

jdarwood007:
If you use the back button in your browser then go forward, all the content is lost.  This is because the information is not sent back to the original prior to unloading the page.  This change simply attaches a event to the unload for window and triggers the base update to occur.  The browser handles the rest if it supports remembering what was entered into the textarea.

Signed-off-by: jdarwood007 <unmonitored+github@sleepycode.com>

### DIFF
--- a/src/lib/SCEditor.js
+++ b/src/lib/SCEditor.js
@@ -622,6 +622,7 @@ export default function SCEditor(original, userOptions) {
 		if (form) {
 			dom.on(form, 'reset', handleFormReset);
 			dom.on(form, 'submit', base.updateOriginal, dom.EVENT_CAPTURE);
+			dom.on(window, 'unload', base.updateOriginal, dom.EVENT_CAPTURE);
 		}
 
 		dom.on(wysiwygBody, 'keypress', handleKeyPress);
@@ -1373,6 +1374,7 @@ export default function SCEditor(original, userOptions) {
 		if (form) {
 			dom.off(form, 'reset', handleFormReset);
 			dom.off(form, 'submit', base.updateOriginal);
+			dom.off(window, 'unload', base.updateOriginal);
 		}
 
 		dom.remove(sourceEditor);


brunoais:
Did you test this?
Why "unload" event and not "beforeunload" event?

jdarwood007:
Yes I did test this fix.  It resolved it for our use case.

Either one should work in theory.  I just tested with unload in my case.

I can reproduce this issue on the https://www.sceditor.com/ home page as well.
1. Load https://www.sceditor.com/
2. Change the content
3. Hit Back in your browser
4. Hit Forward in your browser
5. Content is reset

With this change in place, the changed content should be retained by most browsers.

live627:
`onunload` is responsible for executing an instruction when the page is being closed. It also causes issue with IE and AJAX. 

`onbeforeunload` is more efficient because it does not run in competition with the actual closing of the window and is triggered before `onunload`.

live627 (changes requested):
This only works if the editor is within a form.

src/lib/SCEditor.js
```diff
@@ -622,6 +622,7 @@ export default function SCEditor(original, userOptions) {
 		if (form) {
 			dom.on(form, 'reset', handleFormReset);
 			dom.on(form, 'submit', base.updateOriginal, dom.EVENT_CAPTURE);
+			dom.on(window, 'unload', base.updateOriginal, dom.EVENT_CAPTURE);
```
samclarke, line 625: There's always a textarea so we don't need to check if the editor is within a form.

When testing this I've noticed restoring is not 100% reliable in Chrome with either the `unload` or `beforeunload` events. Chrome seems to, at least sometimes, not change the value of the textarea to the previously saved value until after the editor has been created.

Adding a `pageshow` event that loads the value from the textarea seems to fix the issue as Chrome always appears to have set the textures value to the previously saved one by the time that's triggered. To avoid issues we would need to also use the `pagehide` event to save the value to the textarea.

This seems to be the most reliable from testing so far:
```js
dom.on(window, 'pagehide', base.updateOriginal);
dom.on(window, 'pageshow', handleFormReset);
```

samclarke (changes requested):

live627:
I've noticed that specifying d`autoUpdate: true` will fix this problem. I wonder if that option could use a better name.

samclarke:
`autoUpdate` does the same thing this PR does except more often as it's triggered by the `blur` event.

PR #734, opened 2019-04-23
to.style sometimes undefined copying into editor in copyCSS

MichaelLegg:
We've come across a case where copying formatted text over formatted text in the editor errors with to.style being undefined. This could do with a check to make sure style exists on the 'to' before we try to set anything on it.

### DIFF
--- a/src/lib/dom.js
+++ b/src/lib/dom.js
@@ -757,8 +757,11 @@ export function isInline(elm, includeCodeAsBlock) {
  * @param {HTMLElement} to
  */
 export function copyCSS(from, to) {
-	to.style.cssText = from.style.cssText + to.style.cssText;
+	if (to.style) {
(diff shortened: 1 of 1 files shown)


src/lib/dom.js
```diff
@@ -757,9 +757,12 @@ export function isInline(elm, includeCodeAsBlock) {
  * @param {HTMLElement} to
  */
 export function copyCSS(from, to) {
-	to.style.cssText = from.style.cssText + to.style.cssText;
+	if (to.style) {
+	  to.style.cssText = from.style.cssText + to.style.cssText;
```
brunoais: use tabs only please

brunoais:
I'll try to test it later today

live627:
Looks like this has been addressed in   https://github.com/samclarke/SCEditor/commit/60a33bf17b62a0762fd948f3dcf6b26107dd9e3c

PR #763, opened 2020-01-02
Add missing user input escaping

### LINKED ISSUE #753: XSS Injection
**Expected Behaviour:** No XSS injection allowed.
**Actual Behaviour:** XSS injection.
_Steps to Reproduce Issue:_
1-Click insert an image.
2-On the link copy past this:
``
3-Click insert.

dvz:
Resolves #753

Addressed in [MyBB 1.8.22](https://mybb.com/versions/1.8.22/)

### DIFF
--- a/src/lib/defaultCommands.js
+++ b/src/lib/defaultCommands.js
@@ -1,5 +1,6 @@
 import * as dom from './dom.js';
 import * as utils from './utils.js';
+import * as escape from './escape.js';
 import { ie as IE_VER } from './browser.js';
 import _tmpl from './templates.js';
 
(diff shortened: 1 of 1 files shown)


src/lib/defaultCommands.js
```diff
@@ -494,15 +495,17 @@ var defaultCmds = {
 					var attrs  = '';
 
 					if (width) {
-						attrs += ' width="' + width + '"';
+						attrs += ' width="' + parseInt(width) + '"';
```
euantorano: `parseInt` should be using a base value, see [the MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) for more.

```suggestion
						attrs += ' width="' + parseInt(width, 10) + '"';
```

src/lib/defaultCommands.js
```diff
@@ -494,15 +495,17 @@ var defaultCmds = {
...
 					}
 
 					if (height) {
-						attrs += ' height="' + height + '"';
+						attrs += ' height="' + parseInt(height) + '"';
```
euantorano:
```suggestion
						attrs += ' height="' + parseInt(height, 10) + '"';
```

samclarke:
Sorry for taking so long! Merged, thanks!

PR #791, opened 2021-02-13
Use dompurify internally to prevent XSS

samclarke:
Based off PR #767 this uses dompurify internally to sanitise all HTML.

It would add ~7kb to the compressed size and add dompurify as a dependency but should make all methods of inserting HTML into the editor safe from XSS.

Closes     #767

### DIFF
--- a/src/lib/RangeHelper.js
+++ b/src/lib/RangeHelper.js
@@ -76,7 +76,7 @@ var outerText = function (range, isLeft, length) {
  * @class RangeHelper
  * @name RangeHelper
  */
-export default function RangeHelper(win, d) {
+export default function RangeHelper(win, d, sanitize) {
 	var	_createMarker, _prepareInput,
 		doc          = d || win.contentDocument || win.document,
 		startMarker  = 'sceditor-start-marker',
@@ -112,7 +112,7 @@ export default function RangeHelper(win, d) {
 
 		div           = dom.createElement('p', {}, doc);
 		node          = doc.createDocumentFragment();
-		div.innerHTML = html;
+		div.innerHTML = sanitize(html);
 
 		while (div.firstChild) {
 			dom.appendChild(node, div.firstChild);
--- a/src/lib/SCEditor.js
+++ b/src/lib/SCEditor.js
@@ -380,6 +380,21 @@ export default function SCEditor(original, userOptions) {
 	// Don't deep extend emoticons (fixes #565)
 	base.opts.emoticons = userOptions.emoticons || defaultOptions.emoticons;
 
+	// Allow YouTube iframes
+	// https://github.com/cure53/DOMPurify/issues/340#issuecomment-670758980
+	DOMPurify.addHook('uponSanitizeElement', (node, data) => {
+		if (data.tagName === 'iframe') {
+			var src = node.getAttribute('src') || '';
+			if (src.substr(0, 39) !== 'https://www.youtube-nocookie.com/embed/') {
+				if (node.parentNode) {
+					return node.parentNode.removeChild(node);
+				}
+
+				return null;
+			}
+		}
+	});
+
 	/**
 	 * Sanitize HTML to avoid XSS
 	 *
@@ -388,9 +403,10 @@ export default function SCEditor(original, userOptions) {
 	 * @private
 	 */
 	function sanitize(html) {
-		// This is added as a method so plugins can use it without having to
-		// include their own version of DOMPurify
-		return DOMPurify.sanitize(html);
+		return DOMPurify.sanitize(html,{
+			ADD_TAGS: ['iframe'],
+			ADD_ATTR: ['allowfullscreen', 'frameborder']
+		});
 	};
 
 	/**
--- a/src/lib/SCEditor.js
+++ b/src/lib/SCEditor.js
@@ -396,7 +396,7 @@ export default function SCEditor(original, userOptions) {
 		var allowedUrls = options.allowedIframeUrls;
 
 		if (data.tagName === 'iframe') {
-			var src = node.getAttribute('src') || '';
+			var src = dom.attr(node, 'src') || '';
 
 			for (var i = 0; i < allowedUrls.length; i++) {
 				var url = allowedUrls[i];
@@ -416,6 +416,16 @@ export default function SCEditor(original, userOptions) {
 		}
 	});
 
+	// Convert target attribute into data-sce-target attributes so XHTML format
+	// can allow them
+	domPurify.addHook('afterSanitizeAttributes', function (node) {
+		if ('target' in node) {
+			dom.attr(node, 'data-sce-target', dom.attr(node, 'target'));
+		}
+
+		dom.removeAttr(node, 'target');
+	});
+
 	/**
 	 * Sanitize HTML to avoid XSS
 	 *
@@ -426,7 +436,7 @@ export default function SCEditor(original, userOptions) {
 	function sanitize(html) {
 		return domPurify.sanitize(html, {
 			ADD_TAGS: ['iframe'],
-			ADD_ATTR: ['allowfullscreen', 'frameborder']
+			ADD_ATTR: ['allowfullscreen', 'frameborder', 'target']
 		});
 	};
 
--- a/src/lib/SCEditor.js
+++ b/src/lib/SCEditor.js
@@ -1568,7 +1568,7 @@ export default function SCEditor(original, userOptions) {
 		dom.trigger(editorContainer, 'pasteraw', data);
 
 		if (data.html) {
-			// Sanatise again in case plugins modified the HTML
+			// Sanitize again in case plugins modified the HTML
 			pasteArea.innerHTML = sanitize(data.html);
 
 			// fix any invalid nesting
--- a/src/lib/SCEditor.js
+++ b/src/lib/SCEditor.js
@@ -8,6 +8,7 @@ import _tmpl from './templates.js';
 import * as escape from './escape.js';
 import * as browser from './browser.js';
 import * as emoticons from './emoticons.js';
+import DOMPurify from 'dompurify';
 
 var globalWin  = window;
 var globalDoc  = document;
@@ -380,6 +381,18 @@ export default function SCEditor(original, userOptions) {
 	base.opts.emoticons = userOptions.emoticons || defaultOptions.emoticons;
 
 	/**
+	 * Sanitize HTML to avoid XSS
+	 *
+	 * @param {string} html
+	 * @return {string} html
+	 */
+	base.sanitize = function (html) {
+		// This is added as a method so plugins can use it without having to
+		// include their own version of DOMPurify
+		return DOMPurify.sanitize(html);
+	};
+
+	/**
 	 * Creates the editor iframe and textarea
 	 * @private
 	 */
@@ -558,7 +571,7 @@ export default function SCEditor(original, userOptions) {
 		dom.attr(sourceEditor, 'tabindex', tabIndex);
 		dom.attr(wysiwygEditor, 'tabindex', tabIndex);
 
-		rangeHelper = new RangeHelper(wysiwygWindow);
+		rangeHelper = new RangeHelper(wysiwygWindow, null, base.sanitize);
 
 		// load any textarea value into the editor
 		dom.hide(original);
@@ -1554,7 +1567,7 @@ export default function SCEditor(original, userOptions) {
 		dom.trigger(editorContainer, 'pasteraw', data);
 
 		if (data.html) {
-			pasteArea.innerHTML = data.html;
+			pasteArea.innerHTML = base.sanitize(data.html);
 
 			// fix any invalid nesting
 			dom.fixNesting(pasteArea);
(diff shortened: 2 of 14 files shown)


live627 (changes requested):
This set of changes fixes the injection problem.  Thanks.

I see that you're worried about filesizes like me. I think the cost (7KB) is totality worth it for what the library does.

src/lib/SCEditor.js
```diff
@@ -1554,7 +1567,7 @@ export default function SCEditor(original, userOptions) {
 		dom.trigger(editorContainer, 'pasteraw', data);
 
 		if (data.html) {
-			pasteArea.innerHTML = data.html;
+			pasteArea.innerHTML = base.sanitize(data.html);
```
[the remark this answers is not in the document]
live627: Err, maybe it's for plugins.

src/lib/RangeHelper.js
```diff
@@ -112,7 +112,7 @@ export default function RangeHelper(win, d) {
 
 		div           = dom.createElement('p', {}, doc);
 		node          = doc.createDocumentFragment();
-		div.innerHTML = html;
+		div.innerHTML = sanitize(html);
```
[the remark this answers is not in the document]
samclarke: `endHTML` is appended to `html` if it exists so will be sanitised by this.

src/lib/SCEditor.js
```diff
@@ -379,6 +380,18 @@ export default function SCEditor(original, userOptions) {
 	// Don't deep extend emoticons (fixes #565)
 	base.opts.emoticons = userOptions.emoticons || defaultOptions.emoticons;
 
+	/**
+	 * Sanitize HTML to avoid XSS
+	 *
+	 * @param {string} html
+	 * @return {string} html
+	 */
+	base.sanitize = function (html) {
```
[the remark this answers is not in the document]
samclarke: That would be a good to do if it remains.

While testing this I've realised it might not need to be exposed. If plugins are only ever passed sanitised HTML and all HTML received from plugins is sanitised then only possible reason for a plugin to need to use this would be if it accepted HTML user input directly and inserted it into the DOM directly.

None of the default plugins would need this to be exposed. For now it might make more sense to make this private and expose it later if a use case is found.

src/lib/SCEditor.js
```diff
@@ -379,6 +380,35 @@ export default function SCEditor(original, userOptions) {
 	// Don't deep extend emoticons (fixes #565)
 	base.opts.emoticons = userOptions.emoticons || defaultOptions.emoticons;
 
+	// Allow YouTube iframes
+	// https://github.com/cure53/DOMPurify/issues/340#issuecomment-670758980
+	DOMPurify.addHook('uponSanitizeElement', (node, data) => {
+		if (data.tagName === 'iframe') {
+			var src = node.getAttribute('src') || '';
+			if (src.substr(0, 39) !== 'https://www.youtube-nocookie.com/embed/') {
```
samclarke, line 388: Will need some way to specify allowed iframe URLs.

This will it will require a major version dump as it won't be fully backwards compatible if using any plugins that embed iframes as they will need to be updated to specify allowed URLs.

src/lib/SCEditor.js
```diff
@@ -379,6 +380,56 @@ export default function SCEditor(original, userOptions) {
...
 
+	if (!Array.isArray(options.allowedIframeUrls)) {
+		options.allowedIframeUrls = [];
+	}
+	options.allowedIframeUrls.push('https://www.youtube-nocookie.com/embed/');
+
+	// Create new instance of DOMPurify for each editor instance so can
+	// have different allowed iframe URLs
+	// eslint-disable-next-line new-cap
+	var domPurify = DOMPurify();
```
live627, line 391: Could you expose this via a method? That would enable us to add our own hooks.

src/lib/SCEditor.js
```diff
@@ -379,6 +380,56 @@ export default function SCEditor(original, userOptions) {
...
+	 * Sanitize HTML to avoid XSS
+	 *
+	 * @param {string} html
+	 * @return {string} html
+	 * @private
+	 */
+	function sanitize(html) {
+		return domPurify.sanitize(html, {
+			ADD_TAGS: ['iframe'],
+			ADD_ATTR: ['allowfullscreen', 'frameborder']
```
live627, line 429: Can these be options? Example: We use custom formats for our links which use target=blank.

samclarke: Ideally it should allow everything that's safe by default so configuring wouldn't be needed. I'm surprised dompurify is stripping target=blank, I know it needs rel=noopener to be fully secure in older browsers but even with that it's being stripped.

samclarke: Just realised IE doesn't support rel=noopener so that might explain it

A quick Google suggests adding `rel="noreferrer"` will fix the issue in IE and from testing it appears to work but according to [caniuse](https://caniuse.com/rel-noreferrer) it's only supported in IE 11 in Windows 10 creators update.

Possible solutions I can think of to allow target=blank safely: 

 * Do nothing except require rel=noopener and leave it to the page containing the editor to intercept links and make them safe in IE and Edge. Any page that supports IE and needs the extra security will probably have some way of dealing with this although if that was to strip target=blank from user input it would still be an issue in the editor
 * Add some code to intercept all target=blank links in IE and open them safely using JS 
 * Drop IE WYSIWYG support unless it supports noreferrer (not sure how easy this would be to detect)
 * Drop IE and pre-Blink Edge WYSIWYG support altogether (simplifies the codebase a lot)
 * Don't support target=blank and leave it to formats to add a data attribute specifying if a link is target=blank and adding it back when converting to source

src/lib/SCEditor.js
samclarke: Ideally I'd like to not expose dompurify if possible as it would make it difficult to swap out if ever needed. Is there anything you'd need to add a hook for? If there is then it'll have to be exposed but I'd prefer to avoid it if possible.

src/lib/SCEditor.js
samclarke: I've been thinking about this a bit more. As the point of this is to prevent security issues the first option is probably not a good idea and for compatibility not allowing target=blank wouldn't be good either.

Since this is already going to be a backwards incompatible update it might make sense to just drop IE and legacy Edge WYSIWYG support. A lot of sites seem to be dropping IE support including MS which will be dropping it from 365 and teams in August. IE support will be dropped eventually and doing so now would combine two non-backwards compatible updates in one.

If it was dropped then target=blank can be allowed safely with rel=noopener automatically added if missing. Just depends how important IE WYSIWYG support is.

live627: I looked at the source code for don-purify and saw some es6 features like arrow functions and the spread (lol funny name) operator. The transpile it with babel to support ie. This PR includes it wholesale, so no babel. (Explains the filesize difference between their build and what you reported above)

I'm  glad that Microsoft is going to finally drop IE soon. Disappointed that it took so long. Maybe now any stragglers left be a fire lit under them..

I fully support dropping support for IE and pre-chromium edge. This will help open the door to new  JS features, such as  classes.

Built-in formats do not use target=blank so support for that should be left to a plugin.

samclarke: It's not obvious from the source code but the dompurify NPM package actually contains the transpiled code so this PR shouldn't be including any ES6 features, the exact file this PR should be including is: https://github.com/cure53/DOMPurify/blob/main/dist/purify.es.js

Being able to use modern features, especially default parameters, const/let and arrow functions would definitely be nice. UglifyJS only has limited support for es6 features but I think it does include those.

> Built-in formats do not use target=blank so support for that should be left to a plugin.

If it wasn't such a common attribute I'd agree but as it is and by dropping IE support it can be allowed safely it might be worth including. Still not sure though.

jdarwood007: I wouldn't drop pre-chrome edge.  But IE makes sense.  Lots of enterprises are still on older LTS versions of Windows that have pre-chrome edge.  Some still have IE, but I think its inside that gap of minimal usage over effort required to maintain it.

samclarke: Do you know how long LTS for legacy Edge will last? If it's a long time it might be worth keeping but I can't seem to find anything that says. For regular users support will stop March and I think they will be auto upgraded to the new version in April but I can't find anything about extended support.

I know IE 11 has extended support available which will last until ~2025 but given MS are dropping it from 365 and a few other big sites either have or are going to drop support for it a lot of the web will stop working soon so probably safe to drop that.

jdarwood007: Server 2016, 2019 and Windows LTSC 2015/2016/2019 shipped with legacy edge.  But they can be upgraded to Chrome Edge.
https://support.microsoft.com/en-us/lifecycle/search?alpha=Windows%20Server%202016

2016 Server is support until 2027 it seems.  I would personally target 2025 to 2027 being when dropping pre-chrome edge would be safe as most enterprises will have moved off it... Hopefully.  I wouldn't worry too much about 2019 since it only adds a few more years but the target range of users will be very small.

I wouldn't support IE at all.  It may still ship with Windows, but its disabled by default.  No sensible admin would use IE11 with its lack of proper support for lots of modern tech.  If it works, its a best effort from the very few APIs it has.

samclarke: Is legacy Edge definitely included in LTSC? That page does show an end date for legacy Edge support but doesn't for the versions of IE that are still supported. It seems odd to show an end date for Edge if it is still supported but I can't find anything that says one way or the other.

samclarke: I've just found a [Windows IT Pro Blog post](https://techcommunity.microsoft.com/t5/windows-it-pro-blog/ltsc-what-is-it-and-when-should-it-be-used/ba-p/293181) which says Edge isn't included in Windows 10 Enterprise LTSC so it might not be included in LTSC? I can't tell for sure as it doesn't say about other versions and I can't find anything other documentation about it.

Looking at statcounter usage though it's currently at 0.08% and falling and as it isn't included with Windows 10 Enterprise I'm leaning towards it being safe to drop. It will fallback to a plain a textarea so still usable. If you really need  WYSIYWG support for it I can keep it.

samclarke:
I've tested the plugins and formats and so far everything seems to be working with this. I think all that's left is to decide if to allow target=_blank on links and if DOMPurify needs to be exposed or not.

@live627 is there anything you would need the ability to add hooks to DOMPurify for?

src/lib/SCEditor.js
jdarwood007: Ok, I don't run LTSC servers at my work, so I guess bad info.  I do run 2019 DC servers though and they have Pre-Chrome Edge.  Besides our Remote Desktop servers where we have upgraded them to Chrome Edge.  Not sure how many enterprises do like we do though.

How much effort is it taking to do pre-chrome edge?  It had a lot of HTML5 features we use here.  So from the caniuse, it seems there is no effort to support any version of Edge.  So it seems safe to just drop IE.  If Edge works, it works.

live627:
I'm of the opinion that things should be modular any any fluff optional and not built in. Certainly the youttube command is an extra goodie that should be a plugin . Same for target=_blank. Built in commands don't use it, so adding support for it should not be in the base editor. Gotta be mindful of scope creep and all that.

I don't currently have a need to add a  hook for dompurify. I requested that just in case someone else needs it. I want to preempt a future change to support that. Might as well tackle it now.

src/lib/SCEditor.js
samclarke: The lack of documentation available from Microsoft about legacy Edge support is surprising.

I did find [this page](https://docs.microsoft.com/en-us/microsoft-edge/deploy/about-microsoft-edge) yesterday which says Edge isn't included in LTSB versions, including server 2016, but that's all the official information I can find. I'm a bit confused as I've definitely seen it used on server 2016 but maybe that wasn't the LTS version? As far as I can tell it looks like support might actually end in March for everyone.

Supporting Edge requires a couple of fixes in the code for some quirks around contentEditable and to make it use the fallback clipboard API support. The main issue with it is testing as after IE 11, it has the most problems.

Like I say if you need support for it I can keep it but I'm reluctant to as the market share is very small and it looks like it isn't going to be supported soon based on the little information I can find.

It should still work in source mode if runWithoutWysiwygSupport is enabled and will fallback to a plain textarea if not so won't make anything unusable if not supported.

jdarwood007: Well then, I retract my statement on Edge.  Lets wait until we see reports about support or somebody wants to make the changes to introduce the support.

samclarke:
That's the aim although things like the YouTube command and emoticons support are still included for backwards compatibility as they were included very early on, possibly the first version. At some point they should be moved out though.

I've been thinking about target=blank some more. As it is very common I think the XHTML format should really include support for it but it shouldn't affect BBCode much (at least not in a way that can't be worked around fairly easily) so probably doesn't need to be in the core.

I think it would be better to avoid exposing dompurify if possible so if there isn't a need at the moment I'll keep it private for now. It's easy to make public if ever needed but once exposed it can't be made private again without making a breaking change.

live627 (approved):

PR #794, opened 2021-02-16
ci: add github workflow/action and remove travis ci

live627:
Builds on Travis are subject to a huge queue which adds a delay of at least an hour.


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,34 @@
...
+      run: echo "::set-output name=dir::$(npm root)"
+    - uses: actions/cache@v2
+      id: npm-cache
+      with:
+        path: ${{ steps.npm-cache-dir.outputs.dir }}
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package.json') }}
+        restore-keys: ${{ runner.os }}-node-
+    - if: steps.npm-cache.outputs.cache-hit != 'true'
+      run: npm ci
+    - run: npm install -g grunt-cli
```
samclarke, line 32: Does `npm test` work on GitHub actions? As if it does that would save having to install Grunt CLI globally each time.

I vaguely remember travis had some issue with it a long ago which required installing Grunt globally so just wondering if GitHub actions has the same issue or if this can be removed.

samclarke:
Looks good, thanks! Travis-ci.org is closing soon so this is very useful.

.github/workflows/node.js.yml
jdarwood007: https://docs.github.com/en/actions/guides/building-and-testing-nodejs

From the documentation, they are not installing grunt and running npm test.  So its possible.

live627: Yeah, it does, but in this case is just an alias for `grunt test` https://github.com/samclarke/SCEditor/blob/b63068c41872bb121adac093b82367fc152aec5f/package.json#L42

samclarke: npm adds` node_modules/.bin` to the PATH when executing scripts so as long as GitHub actions isn't changing anything `npm test` should work without having to install grunt globally.

samclarke: Just thought, it might also be worth adding `timeout-minutes` set to something like 5 minutes just in case any code ever contains an infinite loop or something like that.

live627: wut, the default is six hours? That seems excessive. I'll add a limit of ten minutes to mach Travis.

live627:
I had to install `postcss` to get rid of an error 

```
Loading "Gruntfile.js" tasks...ERROR
>> Error: Cannot find module 'postcss'
>> Require stack:
>> - /home/runner/work/SCEditor/SCEditor/node_modules/autoprefixer/lib/utils.js
>> - /home/runner/work/SCEditor/SCEditor/node_modules/autoprefixer/lib/browsers.js
>> - /home/runner/work/SCEditor/SCEditor/node_modules/autoprefixer/lib/autoprefixer.js
>> - /home/runner/work/SCEditor/SCEditor/Gruntfile.js
>> - /home/runner/work/SCEditor/SCEditor/node_modules/grunt/lib/grunt/task.js
>> - /home/runner/work/SCEditor/SCEditor/node_modules/grunt/lib/grunt.js
>> - /home/runner/work/SCEditor/SCEditor/node_modules/grunt-cli/bin/grunt
>> - /home/runner/work/SCEditor/SCEditor/node_modules/grunt/bin/grunt
````

**Comment by samclarke** [comment]
I've done some digging and it seems to be because GitHub actions is using NPM 6 which doesn't automatically install peerDependencies, even if they're included in package-lock.json, whereas NPM 7 does. I thought the entire point of package-lock.json was to avoid issues like this but it seems to version sensitive.

I suspect this could easily cause some subtle, hard to debug issues in the future if the CI is running a different set of packages. There's no guarantee it's even producing the same code now between NPM versions.

There's a bunch of peer dependencies from the various packages:

```
Package                             | peerDependency
------------------------------------+-------------------------------------
grunt-rollup                        | grunt >=0.4.0
autoprefixer                        | postcss ^8.1.0
puppeteer/node_modules/ws           | bufferutil ^4.0.1
puppeteer/node_modules/ws           | utf-8-validate ^5.0.2
@rollup/plugin-node-resolve         | rollup ^1.20.0||^2.0.0
@rollup/pluginutils                 | rollup ^1.20.0||^2.0.0
terser-webpack-plugin               | webpack ^5.1.0
@lodder/grunt-postcss               | grunt >=1.0.4
@lodder/grunt-postcss               | postcss ^8.0.9
grunt-eslint                        | grunt >=1
webpack-dev-middleware              | webpack ^4.0.0 || ^5.0.0
webpack-dev-server                  | webpack ^4.0.0 || ^5.0.0
ajv-keywords                        | ajv ^6.9.1
grunt-contrib-clean                 | grunt >=0.4.5
acorn-jsx                           | acorn ^6.0.0 || ^7.0.0 || ^8.0.0
@babel/helper-compilation-targets   | @babel/core ^7.0.0
grunt-contrib-concat                | grunt >=0.4.0
ajv-errors                          | ajv >=5.0.0
```

Even if `postcss` is the only one not included as a dependency of another package right now, there's no guarantee package updates won't cause similar issues in the future without checking the entire dependency tree each time. It would be fine if NPM 6 and NPM 7 had matching behaviour but as they don't I think the best solution would be to pick one and require that.

As NPM 7 is the latest version it's probably best to use that, plus it works as expected with a package-lock.json from either NPM version. I think either specifying `node_version: 15` or running `npm install -g npm@latest` will give us NPM 7.

While looking into this I noticed there's actions/setup-node@v2 so might be worth upgrading to that.

File: .github/workflows/node.js.yml [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+      run: echo "::set-output name=dir::$(npm root)"
+    - uses: actions/cache@v2
+      id: npm-cache
+      with:
+        path: ${{ steps.npm-cache-dir.outputs.dir }}
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package.json') }}
+        restore-keys: ${{ runner.os }}-node-
+    - if: steps.npm-cache.outputs.cache-hit != 'true'
+      run: npm ci
+    - run: phantomjs --version
```
samclarke, line 33: Also don't need this anymore as `grunt-contrib-qunit` is using headless Chromium instead of PhantomJS now.

.github/workflows/node.js.yml
samclarke: I know, shocked me too when I saw that.

live627:
Huh, I updated  my local npm to v7.6 and it STILL skips peer deps. I'm so frustrated right now.

The build seems ot work if I revert the lockfile

samclarke:
I can see why you're frustrated, it looks very fragile and painful to debug. After looking at it a while I noticed when the postcss dependency was added all the peerDependencies were removed from the lock file so that might explain it.

It appears to be working now though so what do you think? Should we just merge this and hope it keeps working and switch to yarn if it breaks again?

samclarke (approved):

live627:
> After looking at it a while I noticed when the postcss dependency was added all the peerDependencies were removed from the lock file so that might explain it.

I manually reverted those changes which is why it works.

is this why you  originally used yarn? I'm probably going to need to learn how to use it and what the differences are.

npn has never given me any problems before.  also never use Grrunt.

>  Should we just merge this and hope it keeps working and switch to yarn if it breaks again?

I'll be glad to finally put this one behind me. It does `npm ci` rather than `npm install` so it should only read the lockfile. It also stores all dependencies in  a cache, so as to skip downloading dependencies every time. Cache key is based on the contents off the lockfile.

I'll merge this one now.

samclarke:
I originally used yarn as back then it was a to faster and fixed a bunch of other issues (somewhat like this one) with NPM.

NPM has improved a lot since then though. This is the first time I've seen an issue with NPM in a long time, we just seemed to have hit an edge case where two versions treat the lockfile differently.

PR #800, opened 2021-02-22
Drop IE and legacy Edge support

samclarke:
Drop IE and legacy pre-blink Edge support.

This should make it a lot easier to fix bugs / add new featuers as both IE and legacy Edge have a lot of edge cases and require a lot of manual testing. As far as I can tell legacy Edge support is ending in March so should be safe to drop.

Both browser should still work in source mode if `runWithoutWysiwygSupport` is enabled.

### DIFF
--- a/src/formats/bbcode.js
+++ b/src/formats/bbcode.js
@@ -1800,7 +1800,7 @@
 		function convertToHTML(tokens, isRoot) {
 			var	undef, token, bbcode, content, html, needsBlockWrap,
 				blockWrapOpen, isInline, lastChild,
-				ret = [];
+				ret = '';
 
 			isInline = function (bbcode) {
 				return (!bbcode || (bbcode.isHtmlInline !== undef ?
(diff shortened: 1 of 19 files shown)


live627:
also some ie code in the material icon set

samclarke:
Removed. Also found a couple more bits that could be simplified.

src/formats/bbcode.js
```diff
@@ -1800,7 +1800,7 @@
 		function convertToHTML(tokens, isRoot) {
 			var	undef, token, bbcode, content, html, needsBlockWrap,
 				blockWrapOpen, isInline, lastChild,
-				ret = [];
+				ret = '';
```
jdarwood007, line 1804: Do we have to substantiate these as strings in the first place?  Can we just leave them undefined?

jdarwood007:
Overall, I'm shocked at how much code is needed for IE support..

src/formats/bbcode.js
samclarke: Sadly yes, appending a string to an undefined variable causes `undefined` to first be converted to the string  `"undefined"` before appending the other string. E.g.:

```js
var ret;
ret += 'test'; // ret will now equal 'undefinedtest' instead of 'test' as expected
```

We could replace the first concatenation with assignment but given all the if statements it would probably end up harder to read in this case, easier to initialise it to an empty string.

samclarke:
It used to be even worse, this is after a lot of refactoring to reduce it and after dropping IE <= 9.

Hopefully this will make getting all the bugs fixed a bit easier. Using something like [Playwright](https://github.com/microsoft/playwright) it might even be possible to automate testing for all the supported browser engines now.

live627:
> Overall, I'm shocked at how much code is needed for IE support..

I remember needing to write code for IE 6 & 7... was a large time sink. The JavaScript code in SMF is still littered with compatibility code.

live627 (approved):

PR #801, opened 2021-02-22
Update dev dependencies

samclarke:
Updates dev dependencies so everything is up to date.

This removes sauce labs and githook support from Grunt and switches from postcss-clean to to cssnano.

Also switches from yarn.lock to package-lock.json as that's more common now and recent versions of yarn support package-lock.json.

Closes #793, closes #792


live627:
this broke the coverage for some reason (I can't figure it out)

samclarke:
That took longer than expected. The `grunt-contrib-qunit` package switched to headless Chrome which broke coverage. On the plus side I managed to also get rid of the deprecated `istanbul` package.

live627 (approved):

PR #802, opened 2021-02-23
Drop v1 compatibility plugin

samclarke:
Remove the compatibility plugin for version 1 as shouldn't be needed anymore.


live627:
According to my understanding, v1 of SCE relied upon jQuery. Could we then also drop jQuery (seems to be used heavily in the tests)

samclarke:
The only file still using jQuery outside of tests is https://github.com/samclarke/SCEditor/blob/master/src/jquery.sceditor.js which is for backwards compatibility and doesn't have an effect on the rest of the code (it relies on there being a global jQuery variable). That could be deprecated and then dropped though.

jQuery could be removed from the unit tests, would just require replacing the jQuery code.

PR #804, opened 2021-02-24
Prevent copying inherited styles and merge inline styles on paste

### LINKED ISSUE #708: Copy&Paste in Chrome
Hi guys,
Maybe someone knows how to override/solve that problem:
I had to switch to Chrome due to an extension that exist only for that browser, used FireFox before that and never had a problem with Copy&Paste.

Go to sceditor.com, copy the first line, paste it on the second line. Switch to <Source> view and see the result:
In FireFox:
```
Give it a try! :)
Give it a try! :)
```

In Chrome:
```
Give it a try! :)
[color=#111111][size=2][font=Verdana, Arial, Helvetica, sans-serif]Give it a try! [/font][/size][/color]:)
```

Any ideas?
Thanks.

### LINKED ISSUE #764: Randomly adds font or color tags
**Way to reproduce:**
Open the editor, paste text copied from anywhere

**Actual behavior:**
It adds bbcode tags for font color and size automatically. In my case it added `[color=#24292e][size=2][font=-apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji"]`. After several copy and paste (and it can happens very often), the content become a real mess full of undesired bbcode tags.

**Expected behavior:**
The editor shouldn't add tags if not explicitly asked.

samclarke:
This limits what's copied by browsers to exclude inherited styling.

To handle pasting say `<b><em>test</em></b>` into `'<em><b>|</b></em>` which would become `<em><b><b><em>test</em></b></b></em>` it merges a limited set of inline styles and tags so the final HTML will be `<em><b>test</b></em>`.

It will merge inline styles from pasted text if they match the parent tags style and also the b, strong, em, span and font tags when pasting if they match a parent of the same type (both attributes and inline styles).

The exception is `<b>` and `<strong>` which are treated as aliases of each other for merging. Not ideal but if the editor ever ended up with both it would cause a bunch of extra bold tags if they aren't treated as aliases.

Would be good to have some testing around the copy and paste APIs but we would need to start using something like [playwright](https://github.com/microsoft/playwright) first to enable it.

Fixes #708, fixes #764

### DIFF
--- a/src/lib/SCEditor.js
+++ b/src/lib/SCEditor.js
@@ -320,7 +320,7 @@ export default function SCEditor(original, userOptions) {
 		initResize,
 		initEmoticons,
 		handlePasteEvt,
-		handleCopyEvt,
+		handleCutCopyEvt,
 		handlePasteData,
 		handleKeyDown,
 		handleBackSpace,
@@ -664,7 +664,7 @@ export default function SCEditor(original, userOptions) {
 		dom.on(wysiwygBody, 'blur', valueChangedBlur);
 		dom.on(wysiwygBody, 'keyup', valueChangedKeyUp);
 		dom.on(wysiwygBody, 'paste', handlePasteEvt);
-		dom.on(wysiwygBody, 'copy', handleCopyEvt);
+		dom.on(wysiwygBody, 'cut copy', handleCutCopyEvt);
 		dom.on(wysiwygBody, compositionEvents, handleComposition);
 		dom.on(wysiwygBody, checkSelectionEvents, checkSelectionChanged);
 		dom.on(wysiwygBody, eventsToForward, handleEvent);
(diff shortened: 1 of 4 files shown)


samclarke:
Thinking about this some more it might be better to also include the parent styles as well just not the default ones

samclarke:
I'm still debating if this is the best approach but I can't think of a better solution than merging inline tags.

live627:
Your  usage of "default styles" here isn't clear to me. Do you mean browsers impose their own styling to copied text? Maybe my confusion is why I wasn't able to reproduce #708... or I did it wrong.

Anyway, this code seems to make things more reliable or  robust, if that's the right word.

(This gives me some ideas on some more     tests for the bbc parser!)

samclarke:
> Your usage of "default styles" here isn't clear to me.

Yeah that isn't very clear, by default styles I mean those inherited from the browser default styling and stylesheets. I've updated the comments with a better description.

> (This gives me some ideas on some more tests for the bbc parser!)

That sounds good! I'm currently trying out playwright and so far it's looking promising. I'm really hoping it will allow testing things like this as that would allow for much better test coverage.

PR #805, opened 2021-02-26
Specify base when parsing integers

euantorano:
When using `parseInt` to resolve image width/height, [MDN recommends](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#description) passing an explicit `base` as the second parameter.


samclarke:
Looks good, thanks! Merged.

PR #806, opened 2021-03-04
fix: Set focus to the editor after state change

live627:
Keyboard shortcuts only apply when thee editor is in focus, so state changes (toggling editor modes, maximizing) should set focus to the editor.

This unexpected behavior was reported at https://github.com/SimpleMachines/SMF2.1/issues/4698


samclarke (approved):

samclarke:
Merged, thanks!

PR #807, opened 2021-03-05
Append a zero width space to inserted links

### LINKED ISSUE #644: Bug - Firefox only: Insert a Link (Hyperlink) doesn't stop editing of the link it self
Insert a Link (Hyperlink) doesn't stop editing of the link it self
Firefox only.  (Tested with version: 58.0)

1. Remove the content of the editor
2. Type in some text like "abc"
3. Click on 'Insert a link' and type in some text like "http://example.com", leave description empty
4. Click on "Insert" to insert the link
4. DON'T CLICK ANYWHERE ELSE
5. Just continue typing with the text "def".

=> Result:
abc[url=http://example.com]http://example.comdef[/url]

=> Expected result:
abc[url=http://example.com]http://example.com[/url]def

Works perfectly in chrome!

live627:
Fixes #644

### DIFF
--- a/src/lib/defaultCommands.js
+++ b/src/lib/defaultCommands.js
@@ -544,11 +544,15 @@ var defaultCmds = {
 						editor.wysiwygEditorInsertHtml(
 							'<a href="' +
 							'mailto:' + escape.entities(email) + '">' +
-								escape.entities((text || email)) +
-							'</a>'
+								escape.entities(text || email) +
+							'</a>&#8203;'
 						);
 					} else {
-						editor.execCommand('createlink', 'mailto:' + email);
+						editor.wysiwygEditorInsertHtml(
+							'<a href="' +
+							'mailto:' + escape.entities(email) + '">',
+							'</a>&#8203;'
+						);
 					}
 				}
 			);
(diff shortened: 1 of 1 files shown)


brunoais:
What happens with the BBCode plugin and the xHTML plugin when they encounter that character?
Is a filter necessary to avoid sending those characters when trying to get their value?

samclarke:
Adding a zero width space will cause backspace to not behave as expected. It might be better after inserting the HTML to move the cursor if it's inside the link. Something like:

```js
// Set cursor to after the a element if A is the parent
function placeCursorAfterLink(rangeHelper) {
	var parent = rangeHelper.parentNode();
	if (parent && parent.nodeName === 'A') {
		var range = rangeHelper.selectedRange();
		range.setStartAfter(parent);
		rangeHelper.selectRange(range);
	}
}

// After inserting link HTML move the cursor
placeCursorAfterLink(editor.getRangeHelper());
```

@brunoais you're right it would be best to avoid including any formatting in the final output. I think in this case the `removeWhiteSpace()` function will already strip any zero width spaces so it should be OK.

live627:
That works... until you click the editor and notice that the link is still selected when you keep typing.

jdarwood007:
I agree with Sam's approach.

brunoais:
@samclarke I think that should only happen if the user places the cursor after the last character of the link.

IMO:
When the user clicks at the end of the link, the cursor should be placed after it (and not when clicking the link). If then the user presses left, the cursor should go inside the link and after the last character (the link button becoming active should be the feedback). To exit the anchor, just press arrow to the right until 1ce after the last character of the link.

PR #808, opened 2021-03-06
feat:Support excluded html tag attributes in bbcode formats with `false`

live627:


live627:
I notice that this function assumes that returning false short circuits utilis.each, similar to dom.traverse. Only... it doesn't.  Was that an oversight or was each()never intended to short circuit?

src/formats/bbcode.js
```diff
@@ -539,23 +537,20 @@
 			allowsEmpty: true,
 			tags: {
 				img: {
-					src: null
+					src: null,
+					[EMOTICON_DATA_ATTR]: false
```
brunoais, line 541: Which browsers support this variable key notation?

live627: [Is this what you're looking for?](https://caniuse.com/mdn-javascript_operators_object_initializer_computed_property_names)

live627:
Digging deeper to answer my own question, it seems that the code in question is a holdover from back when jQuery was used. I'll need to test it to see how it breaks things.

src/formats/bbcode.js
brunoais: Yes it is.

live627:
I'm converting this one to a draft for now until I can write up a bunch of unit tests for the possibility of compound tag matching.

live627:
Before i look into writing some tests for this, what are your thoughts @samclarke

samclarke:
Looks good, would be useful to have. It does look like it would change the behaviour of `false` in matching but as only `null` is documented that should be fine.

live627:
I actually didn't think of that. I'm going to set this to target v4.0.0 because of the BC break. Might be silly since it is undocumented. And there may be a chance that someone might be doing it that way.

PR #820, opened 2021-04-12
Improve fixNesting()

### LINKED ISSUE #545: Inserting hr sometimes splits quotes in chrome
When inserting a `<hr />` between lines inside a quote it sometimes splits the quote in chrome.

Source: https://community.mybb.com/thread-173071.html

### LINKED ISSUE #768: [code] tag gets reproduced on every new line
Assuming this project is dead and no one is maintaining it. It will be worth a try at least to report a bug.

When marking text and clicking the code button, then [code] gets printed on every single line rather than the beginning and the end of the block.

    $('textarea[name="message"]').sceditor({
      format: "bbcode",
      plugins: "plaintext",
      pastetext: {
        addButton: true,
        enabled: true,
      },
      style: "/vendor/sceditor/themes/content/default.min.css",
      locale: "en",
      emoticonsEnabled: false,
      enablePasteFiltering: true,
      toolbar: "bold,italic,underline,strike|removeformat|bulletlist,orderedlist,table|code,quote|horizontalrule,image,link,unlink|source",
      bbcodeTrim : true,
      parserOptions: {
        breakBeforeBlock: false,
        breakStartBlock: false
      },
      width: "auto"
    });

image

### LINKED ISSUE #815: fixNesting() doesn't fix <p> containing a <blockquote>
The `fixNesting()` function doesn't correctly handle `<p>` tags containing a `<blockquote>`. It should split any `<p>` tags that contain a `<blockquote>` but it currently does nothing.

For example, calling `fixNesting()` with:

```html
<p>1<blockquote>2</blockquote>3</p>
```

should be converted to:

```html
<p>1</p><blockquote>2</blockquote><p>3</p>
```

The browser will fix the invalid nesting when HTML is set via `innerHTML` which can then cause an issue with extra lines being added to the output.

What happens is when the contents is converted, the browser fixes the invalid nesting which can cause empty `<p>` tags to be inserted around the `<blockquote>` which can then result in unexpected new lines being added to the output.

Fixing this might also fix #545.

samclarke:
Updates `fixNesting()` to correctly fix nesting for `<p>` tags containing block elements. Fixes #815, fixes #545 and fixes #768

Changes the behaviour of `fixNesting()` to not split inline elements containing comments.

Also prevents fixNesting() from creating empty tags and copies inline styling when fixing nesting, both of which make more sense and should result in a better UX when editing.

Deprecates the `copyCSS()` function as no longer needed by `fixNesting()`. Should probably also deprecate `findCommonAncestor()` as that's no longer needed and can be removed at some point too.

### DIFF
--- a/src/lib/dom.js
+++ b/src/lib/dom.js
@@ -779,6 +779,14 @@ export function fixNesting(node) {
 		return node;
 	};
 
+	function isEmpty(node) {
+		if (node.lastChild && isEmpty(node.lastChild)) {
+			remove(node.lastChild);
+		}
+
+		return node.nodeType === 3 ? !node.nodeValue : !node.childNodes.length;
+	}
+
 	traverse(node, function (node) {
 		var list = 'ul,ol',
 			isBlock = !isInline(node, true),
@@ -797,8 +805,13 @@ export function fixNesting(node) {
 			// it still has the same styling
 			copyCSS(parent, middle);
 
-			insertBefore(before, parent);
 			insertBefore(middle, parent);
+			if (!isEmpty(before)) {
+				insertBefore(before, middle);
+			}
+			if (isEmpty(parent)) {
+				remove(parent);
+			}
 		}
 
 		// Fix invalid nested lists which should be wrapped in an li tag
--- a/src/lib/dom.js
+++ b/src/lib/dom.js
@@ -757,7 +757,7 @@ export function isInline(elm, includeCodeAsBlock) {
  * @param {HTMLElement} to
  */
 export function copyCSS(from, to) {
-	if (to.style && from.style) {
+	if (to.style && from.style && from.style.cssText) {
 		to.style.cssText = from.style.cssText + to.style.cssText;
 	}
 }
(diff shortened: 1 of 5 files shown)


src/lib/dom.js
```diff
@@ -755,6 +755,7 @@ export function isInline(elm, includeCodeAsBlock) {
  *
  * @param {HTMLElement} from
  * @param {HTMLElement} to
+ * @deprecated since v3.1.0
  */
 export function copyCSS(from, to) {
```
live627, line 760: Should a warning be logged to the console if someone's code calls this?

samclarke: That's a great idea!

live627 (approved):
Were you going to add a warning if someone's code calls `copyCSS()`?

live627:
I'm going to merge this in now in the interest of moving development forward. There's still the question of adding a warning if someone's code calls copyCSS()

live627:
@samclarke I've noticed that a review is required for merging in pull requests now. The timing is a bit sus and seems to indicate that you didn't want me to merge this latest batch of PRs.

samclarke:
@live627 One of the merged PR's had changes requested, so I've temporarily required all PR's to be reviewed while I review that one.

Issue #589, opened 2017-11-28
Improve HTML to BBCode conversion code


samclarke:
Investigate updating BBCode `tags` to support style selection and re-map `styles` to be part of `*` tag selector. Can then simplify the conversion code to just call the `handleTags()` function and possibly also deprecate the `styles`.

From #586.

PR #821, opened 2021-04-13
 Add ability to match all attrs in a given tag 

### LINKED ISSUE #809: Matching multiple attrs is very broken
**Describe the bug**
This bug is two-fold:
- fI the value of the attr does not match one of the strings in the array,  `convertBBCode` is not set to false`. This behavior is reproduced by reversing the order of the attrs below.
- If the final attr matches successfully, `convertBBCode` becomes `true` without considering that otter attrs may fail. The  below example reproduces this one.
None of the built in formats have multiple attrs and is why this is a seemingly hidden bug. (The most egregious ones seem to always hide in plain sight!)

I have a fix for this but still need to write a bunch of tests to prove that  it works with all permutations.

**To Reproduce**

			sceditor.formats.bbcode.set(
				'abbr', {
					tags: {
						abbr: {
							title: ['test'],
							test: null
						}
					},
					format: function (element, content) {
						return '[abbr=' + element.getAttribute('title') + ']' + content + '[/abbr]';
					},
					html: '<abbr title="{defaultattr}">{0}</abbr>'
				}
			);
			var textarea = document.getElementById('example');
			sceditor.create(textarea, {
				format: 'bbcode',
				icons: 'monocons',
				style: '../minified/themes/content/default.min.css'
			});
			var parser = sceditor.instance(document.getElementById('example'));
			console.log(parser.toBBCode(
				'<abbr title="attrs.defaultattr" test="value">content</abbr>'
			));

**Expected output**
`content`

**Actual output**
`[abbr=attrs.defaultattr]content[/abbr]`

live627:
This is my set of changes to add an additional mode for matching attributes on HTML tags—require all to match. The default is to match any.

Fix #809

New option added `instance.opts.parserOptions.matchAttrs`
New enumerable added: `sceditor.BBCodeParser.MatchingMode`

The enum has two values:

- `sceditor.BBCodeParser.MatchingMode.all`
- `sceditor.BBCodeParser.MatchingMode.any`—this is the value which will be used if left blank, ie, not set. This way, any existing code will still work exactly as before.

Each individual format option will have the opportunity to specify its own `matchAttrs` and override the global option.

### DIFF
--- a/src/formats/bbcode.js
+++ b/src/formats/bbcode.js
@@ -50,6 +50,21 @@
 		auto: 3
 	};
 
+	var MatchingMode = {
+		/** @lends BBCodeParser.MatchingMode */
+		/**
+		 * Match all values.
+		 * @type {Number}
+		 */
+		all: 1,
+
+		/**
+		 * Match any value. This is the default if unspecified.
+		 * @type {Number}
+		 */
+		any: 2
+	};
+
 	var defaultCommandsOverrides = {
 		bold: {
 			txtExec: ['[b]', '[/b]']
@@ -2084,9 +2099,22 @@
 	 * @type {Object}
 	 * @class QuoteType
 	 * @name BBCodeParser.QuoteType
-	 * @since v1.4.0
+	 * @since 1.4.0
 	 */
 	BBCodeParser.QuoteType = QuoteType;
+
+	/**
+	 * Method to use when matching attributes.
+	 *
+	 * - MatchingMode.all roughly equates to Array.every().
+	 * - MatchingMode.any roughly equates to Array.some().
+	 *
+	 * @type {Object}
+	 * @class MatchingMode
+	 * @name BBCodeParser.MatchingMode
+	 * @since 3.1.0
+	 */
+	BBCodeParser.MatchingMode = MatchingMode;
 
 	/**
 	 * Default BBCode parser options
@@ -2151,7 +2179,21 @@
 		 * @type {BBCodeParser.QuoteType}
 		 * @since 1.4.1
 		 */
-		quoteType: QuoteType.auto
+		quoteType: QuoteType.auto,
+
+		/**
+		 * Method to use when matching attributes.
+		 *
+		 * - MatchingMode.all roughly equates to Array.every().
+		 * - MatchingMode.any roughly equates to Array.some().
+		 *
+		 * This will be used as a fallback on tags that do not
+		 * specify a matching mode.
+		 *
+		 * @type {BBCodeParser.MatchingMode}
+		 * @since 3.1.0
+		 */
+		matchAttrs: MatchingMode.any
 	};
 
 	/**
@@ -2226,7 +2268,7 @@
 		 * Same as tagsToBBCodes but instead of HTML tags it's styles
 		 * @private
 		 */
-		var stylesToBBCodes = {};
+		var stylesToBBCodes = [];
 
 		/**
 		 * Allowed children of specific HTML tags. Empty array if no
@@ -2242,40 +2284,35 @@
 		};
 
 		/**
-		 * Populates tagsToBBCodes and stylesToBBCodes to enable faster lookups
+		 * Populates tagsToBBCodes and stylesToBBCodes for easier lookups
 		 *
 		 * @private
 		 */
 		function buildBbcodeCache() {
-			each(bbcodeHandlers, function (bbcode) {
-				var	isBlock,
+			each(bbcodeHandlers, function (bbcode, handler) {
+				var
 					tags   = bbcodeHandlers[bbcode].tags,
-					styles = bbcodeHandlers[bbcode].styles;
+					styles = bbcodeHandlers[bbcode].styles,
+					// TODO: Move this to utils
+					isObject = function (o) {
+						return o instanceof Object && o.constructor === Object;
+					};
 
 				if (tags) {
 					each(tags, function (tag, values) {
-						isBlock = bbcodeHandlers[bbcode].isInline === false;
-
-						tagsToBBCodes[tag] = tagsToBBCodes[tag] || {};
-
-						tagsToBBCodes[tag][isBlock] =
-							tagsToBBCodes[tag][isBlock] || {};
-
-						tagsToBBCodes[tag][isBlock][bbcode] = values;
+						tagsToBBCodes[tag] = tagsToBBCodes[tag] || [];
+						tagsToBBCodes[tag].push([
+							isObject(values) ? Object.entries(values) : values,
+							handler.matchAttrs || base.opts.matchAttrs || 2,
+							handler.format,
+							handler.isInline === false
+						]);
 					});
 				}
 
 				if (styles) {
 					each(styles, function (style, values) {
-						isBlock = bbcodeHandlers[bbcode].isInline === false;
-
-						stylesToBBCodes[isBlock] =
-							stylesToBBCodes[isBlock] || {};
-
-						stylesToBBCodes[isBlock][style] =
-							stylesToBBCodes[isBlock][style] || {};
-
-						stylesToBBCodes[isBlock][style][bbcode] = values;
+						stylesToBBCodes.push([style, values, handler]);
 					});
 				}
 			});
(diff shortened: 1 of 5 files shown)


tests/unit/formats/bbcode.parser.js
```diff
@@ -872,11 +872,13 @@ QUnit.test('Code', function (assert) {
 		'Normal'
 	);
 
+	/* TODO: Find out why this assertion fails
 	assert.htmlEqual(
 		this.parser.toHTML('[code]Testing [b]test[/b][/code]'),
 		'<code>Testing [b]test[/b]<br /></code>',
 		'Normal'
 	);
+	*/
```
live627, line 881: Strangely enough, this test fails where `[b]` is converted to `<strong>`, even though similar code behaves on both Firefox and Chrome or me.

			var parser = sceditor.instance(document.getElementById('example'));
			console.log(parser.fromBBCode(
				'[code]Testing [b]test[/b][/code]'
			));

i''m guessing this may be why `handleTags()` differentiated between block and inline.

Also, the tests are a pain to debug because not only the line numbers are messed up (yay for monolithic files) but `console.log()` doesn't work.

live627: ok so  I reverted that one and surprisingly the assertion passes here. Dunno why. Not gonna look into it because this test suite is a royal pain for me to debug.

I wonder if it's because I use Windows.

live627: And now after adding the latest gamut of tests, the problematic one passes on my end. Amusing.

Seems like a consequence of global state.

src/formats/bbcode.js
```diff
@@ -2242,40 +2284,33 @@
...
 		function buildBbcodeCache() {
-			each(bbcodeHandlers, function (bbcode) {
-				var	isBlock,
+			each(bbcodeHandlers, function (bbcode, handler) {
+				var
 					tags   = bbcodeHandlers[bbcode].tags,
-					styles = bbcodeHandlers[bbcode].styles;
+					styles = bbcodeHandlers[bbcode].styles,
+					// TODO: Move this to utils
+					isObject = o => o instanceof Object
```
samclarke, line 2297: For now the JS needs to remain compatible with IE and other old browsers so the editor can run in source mode when `runWithoutWysiwygSupport` is set.

I should have said that when I deprecated IE support., it's why I haven't updated the code to using the modern features yet, no matter how much I really want to.

This would either have to wait until the next breaking release or be converted to ES5 (urgh).

src/formats/bbcode.js
```diff
@@ -2505,15 +2507,10 @@
...
-								// handle inline bbcodes
 								curTag = handleStyles(node, curTag);
-								curTag = handleTags(node, curTag);
-
-								// handle blocklevel bbcodes
-								curTag = handleStyles(node, curTag, true);
 							}
 
-							curTag = handleTags(node, curTag, true);
+							curTag = handleTags(node, curTag);
```
samclarke, line 2513: Inlines need to be converted before blocks otherwise it can cause issues with blocks being nested inside of inlines.

I'm not sure if the issue exists with the default BBCodes but for example if you rename the `b` BBCode to `z`:

```js
sceditor.formats.bbcode.set('z', sceditor.formats.bbcode.get('b'));
sceditor.formats.bbcode.remove('b');
```

then create the editor and convert:

```html
<div style="font-weight: bold; text-align: center;">test</div>
```

it will be converted to `[b][center]test[/center][/b]` instead of `[center][b]test[/b][/center]` which it currently does. Thinking about it, there should really be a test to cover this.

src/formats/bbcode.js
```diff
@@ -50,6 +50,21 @@
 		auto: 3
 	};
 
+	var MatchingMode = {
```
samclarke, line 53: I'm wondering about this, do you think there will ever be a need for anything other than `all` and `any` matching? I'm thinking anything else is likely to be very complex and require a different API.

If not, using a flag like `strictMatching: true` (not sure about the name) might be simpler. Or if you think there might be more matching types then using strings might be better.

I know `quoteType` does it like this but I'm not a massive fan of that either. It was an attempt to improve autocompletion but it didn't really help and ended up not being very JS like. I've been wanting to replace it with plain strings at some point as, if typescript definitions are added, the autocompletion will work just as well if not better and would be more JS like. What do you think?

samclarke (changes requested):
Sorry for taking so long to review this, I've been a bit busy. Looks like it will be a really useful feature!

tests/unit/formats/bbcode.parser.js
samclarke: That's really strange, I might try an investigate the cause of this as it looks like a bug somewhere in the code.

live627: My workaround is https://github.com/samclarke/SCEditor/pull/821/commits/81a56bbadc5c7ab2db8f512de1600dbc3cea4c5f

Seems to be global state. I'm leaving it be since it works.

src/formats/bbcode.js
live627: My goal with redoing this code was to make it more concise and easy to read. The double calls for blocks and inlines annoyed me,. That bug looks to be fixed by separating the styes back out and not the tags https://github.com/samclarke/SCEditor/pull/821/commits/58fab8598aa92ece8db59c9ed01813166dd21dbe

src/formats/bbcode.js
live627: My view on the enums is that they help make the public API easier to use and understand. That said, a simple bool here could suffice given a descriptive name. Maybe `matchAllAttrs`. I still think that it may not be as meaningful as a full enum. I'm kinda torn on this.

src/formats/bbcode.js
live627: I have added tests to ensure proper nesting for both styles and tags.

src/formats/bbcode.js
live627: I just thought of a possibility to add two new possible types
-   `ArrayList` or a`SimpleArray`
- `Callback`

src/formats/bbcode.js
samclarke: The duplicate calls annoy me too. One thing I wanted to do at some point to simplify the code is move styles in as an attribute (#589) so defining a BBCode would be:

```js
// Match any tag with text-align (equivalent to current styles behaviour)
'*' {
  style: {
    'text-align': null
  }
}
```
or
```js
// Match any p tag with text-align (currently requires matching one or the other and
// manually checking in the format function)
'p' {
  style: {
    'text-align': null
  }
}
```

That way only `handleTags` has to be called twice and `handleStyles` gets remove completely. `handleTags` tags would need to special case the `style` attribute and check for a match in whatever the current matching mode is but even so, that should simplify things. It would also be backwards compatible as we can simply move styles to a * tag as above.

Thinking about it, that might be worth doing for this as currently match all will either matches all styles or all the tags, not both. E.g.:

```js
sceditor.formats.bbcode.set('margin', {
	tags: {
		'p': null
	},
	styles: {
		'margin': null
	},
	isInline: false,
	matchAttrs: sceditor.BBCodeParser.MatchingMode.all,
	format: function (element, content) {
		return '[margin=' + element.style.margin + ']' +
			content + '[/margin]';
	},
	html: '<p style="margin:{defaultattr}">{0}</p>'
});
```

With:

```html
<p>No match</p>
<div style="margin: 1em;">No match</div>
<p style="margin: 1em;">Match</p>
```

Produces:

```
[margin=0px]No match[/margin]
[margin=14px]No match[/margin]
[margin=14px][margin=14px]Match[/margin]
[/margin]
```

If it was done then potentially there could even be multiple matching modes:

`'any'` (current)
`'all-attrs'` (all attributes, any styles)
`'all-styles'` (any attributes, all styles)
`'all'` (all styles and all attributes)

Not sure if that's a good idea as it does look a bit confusing. I definitely think being able to match the style with the tag would be better though.

src/formats/bbcode.js
samclarke: What would ArrayList / SimpleArray do? Callback could be useful although that can be done inside of `format` and if not, just setting a function and having `handleTags` check for it would likely be safer than relying on matching mode being set..

If there can be multiple modes, just using strings would probably be more JS like, it's what I want to do with `quoteType` as that ended up being clunky to type out without any real benefits.

tests/unit/formats/bbcode.parser.js
samclarke: Oh I see, it's a bug with one of the tests replacing the test parser instance :+1:

src/formats/bbcode.js
live627: I'll make this a string then as it seems that strings are used everywhere like you said. Enums seem to be an anomaly in JS (but are common in compiled languages such as C#)

`array` would just be an array like that in the `styles` attributive and would match with `Array. prototype.includes()`.

live627: Ehh, maybe not. Magic string seem too highly error prone due to typos.

src/formats/bbcode.js
live627: I definitely had #589 in mind when I simplified the data structures. And since `matchAttrs` only affects attributes, your example stays the same regardless of the matching mode.

Matching modes for styles should perhaps be its own property, `matchStyles`, something like that. Matching all tags would probably require some code refactoring as the current setup doesn't look      like it'll facilitate that well. Both features should be separate I think.

Putting styles under tags should be its own PR. It's out of scope of this one.

live627:
I'm going to merge this in now in the interest of moving development forward.

PR #828, opened 2021-04-29
Fix extend() deep merging mixed types

### LINKED ISSUE #827: extend9(() can sometimes try to write to a string as if it were an object, throwing an error
There is an error with extend() when setting up advanced emoticon objects. The problem is caused when extend() tries to write to a string as if  it were an object with keys.

I've distilled this down into a simple test

```
QUnit.test('extend() - deep string', function (assert) {
	var a = { a: {'foo': 'bar' } };
	var b = { a: {'foo': { 'bar': 'baz' } } };

	assert.deepEqual(utils.extend(true, {}, a, b), { a: {'foo': { 'bar': 'baz' } } });
});
```

samclarke:
Should only merge if both source and target tyes match. Fixes #827

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -57,27 +57,37 @@ export function extend(targetArg, sourceArg) {
 	var target = isTargetBoolean ? sourceArg : targetArg;
 	var isDeep = isTargetBoolean ? targetArg : false;
 
+	function isObject(value) {
+		return value !== null && typeof value === 'object' &&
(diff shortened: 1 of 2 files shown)


live627 (approved):

live627:
Oh sorry, I didn't check the assigned milestone before merging this.

samclarke:
My mistake, I selected the wrong milestone I meant to put v3.1.0.

PR #832, opened 2021-05-05
feat: Linebreak in toolbar

### LINKED ISSUE #731: Linebreak in toolbar
Hello,

is it possible to add a newline in the toolbar ?

So i got the following example Toolbar:
```
font,size,color|
bold,italic,underline|left,center,right
```
Instead of
```
font,size,color|italic,underline|left,center,right
```
?

live627:
Use a double pipe `||` to add a new row

Closes #731

### DIFF
--- a/src/lib/SCEditor.js
+++ b/src/lib/SCEditor.js
@@ -1,4 +1,4 @@
-﻿import * as dom from './dom.js';
+import * as dom from './dom.js';
 import * as utils from './utils.js';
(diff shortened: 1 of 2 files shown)


live627:
@samclarke this is completely backwards compatible so could it go into v3.1.0? Was assigning it to v4.0.0 a mistake?

live627:
Can we get this into v3.3.0?

PR #853, opened 2021-10-10
Improve insertnode

samclarke:
Will move inserted inline nodes at the start of end or the input into the start/end containers.

For example, Inserting:
```html
<em>foo</em>
```
into:
```html
<b>|text</b><u>text|</u>
```
will now become:
```html
<b><em>foo</em></b>
```
instead of what it currently becomes:
```html
<em>foo</em>
```


live627:
How would I test this?

PR #879, opened 2022-02-07
missing some phrases in ru.js

zebroid:
missing some phrases in ru.js


samclarke:
Merged. Thanks for translating!

PR #882, opened 2022-03-10
Finnish translation

RJuho:
Hei!

Greeting from Finland, here you have Finnish translation.
Keep going good work!


samclarke:
Looks good. Thanks for translating!

PR #894, opened 2022-08-11
Initialize link dropdown description

JeffreyOfYOSPOS:
At present, when you select text and use the "link" command, a dropdown menu appears for the url itself and the description. However, the initial selected text is not propagated to the "description" box, which is confusing. Populate this box with any text that was selected before the url button was hit.

### DIFF
--- a/src/lib/defaultCommands.js
+++ b/src/lib/defaultCommands.js
@@ -563,13 +563,14 @@ var defaultCmds = {
 
 	// START_COMMAND: Link
 	link: {
-		_dropDown: function (editor, caller, cb) {
+		_dropDown: function (editor, caller, selected, cb) {
 			var content = dom.createElement('div');
 
 			dom.appendChild(content, _tmpl('link', {
 				url: editor._('URL:'),
 				desc: editor._('Description (optional):'),
-				ins: editor._('Insert')
+				ins: editor._('Insert'),
+				content: selected
 			}, true));
 
 			var linkInput = dom.find(content, '#link')[0];
--- a/src/formats/bbcode.js
+++ b/src/formats/bbcode.js
@@ -206,6 +206,7 @@
 				getEditorCommand('link')._dropDown(
 					editor,
 					caller,
+					selected,
 					function (url, text) {
 						editor.insertText(
 							'[url=' + url + ']' +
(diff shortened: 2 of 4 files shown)


samclarke:
Thanks for the PR!

This would be a much nicer UX for source mode but it would result in a different UX for WYSIWYG and source mode. I'm not sure if that's a good idea or not.

It's not really possible to show the selection in the description when in WYSIWYG mode as it could contain an image or something else that can't be represented in text.

Could possibly show a placeholder like `[current selection]` when in WYSIWYG mode although I'm not sure if that's better or not. Another possibility would be to remove the description field entirely and require changing the link text after inserting.

src/lib/defaultCommands.js
```diff
@@ -563,13 +563,14 @@ var defaultCmds = {
 
 	// START_COMMAND: Link
 	link: {
-		_dropDown: function (editor, caller, cb) {
+		_dropDown: function (editor, caller, selected, cb) {
```
samclarke, line 566: This causes an error in WYSIWYG mode as the call to `_dropDown` in `exec:` hasn't been updated to pass the `selected` parameter so `cb` is is being treated as `selected` and there is no `cb` passed.

samclarke (changes requested):
Look good! Just missed one of the calls to `_dropDown`.

JeffreyOfYOSPOS:
Thank you for the review! I'm sorry I did not catch that one issue - I will fix it.

I totally understand if you don't like the asymmetry with wysiwig mode. Something Awful is using SCEditor for the buttons primarily and I'm gonna keep thinking about how to make those work better, but I will do my best to test with wysiwig mode anything that I push upstream. I could see removing the description field in wysiwig mode exclusively, but, in my opinion, it  is very important in source mode -it simultaneously simplifies the process of making a adding a link to a post, and also teaches new users the syntax for doing that directly.

A version where selecting stuff in wysiwig mode produced a placeholder would be neat for sure but also a much larger project than simply populating a text input with plaintext. 

I would like it if we also used wysiwig mode some day but our set of accepted bbcode is large and weird and we'd basically end up writing a full bbcode parser in javascript. We have lower hanging fruit than that.

PR #902, opened 2022-11-08
Reduce _normaliseColour regular expression range

samclarke:
This reduces the regular expression range in the _normaliseColour function. The regular expression should only match `0-9a-f` but it currently matches `0-f` which includes some extra characters.

This shouldn't matter too much as the function only attempts to normalise colours and returns the string as is if it can't but it should leave strings that don't match the `#xxx` format alone.

### DIFF
--- a/src/formats/bbcode.js
+++ b/src/formats/bbcode.js
@@ -2188,6 +2188,7 @@
 
 		return number.length < 2 ? '0' + number : number;
 	}
+
 	/**
 	 * Normalises a CSS colour to hex #xxxxxx format
 	 *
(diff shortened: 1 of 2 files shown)


live627:
> it currently matches `0-f` which includes some extra characters.

so was it erroneously matching the ascii range between `9` and `a`?

samclarke:
Exactly. It would treat strings with characters between `0-a` as valid `#xxx` and duplicate them out to `#xxxxxx`.

This makes it leave those invalid strings alone and behave more as expected.

live627 (approved):

PR #905, opened 2022-11-08
Update node action and remove cache action

samclarke:
Updates the actions to the latest versions, updates node and removes the cache action as the setup-node action now has built in cache support.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -15,20 +15,11 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-node@v2
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
       with:
-        node-version: 15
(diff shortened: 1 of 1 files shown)


live627:
This appears to change the cache dir from `node _modules` to `.npm`.

samclarke:
Yeah I _think_ that's what it does. I did initially update cache-action but the docs recommended against caching `node_modules` and then I discovered setup-node has caching built in so figured might as well use that.

I think the reasons they recommend for not caching `node_modules` would be unlikely to occur but as far as I can tell the speed difference is minimal. Plus, these are run so infrequently that half the time the cache has expired anyway.

live627:
it would skip `npm ci` if the cache was found.

live627 (approved):

samclarke:
The difference includes skipping `npm ci`. Loading from the cache seems to be a little slow with GitHub actions which makes the difference fairly small (cache action takes ~6s, `npm ci` using built in cache option takes ~9s and `npm ci` without cache takes ~22s).

PR #935, opened 2023-04-06
Fix checkWhitespace() error when no previous node

### LINKED ISSUE #933: Smiley at the beginning + emoticonsCompat: true -> Uncaught TypeError: prev is null
Firefox 108
SCEditor 3.2.0  `format: bbcode`
The options are set `emoticonsCompat: true`

1. An empty edit box (either immediately after page load (not 100% sure here) or after clearing Ctrl+A and Delete)
2. I insert a smilie through the smilies panel
   
sm1

3. I'm typing

sm2
Recurring `TypeError: prev is null`

sm3

samclarke:
Fixes error with `checkWhitespace()` when using `emoticonsCompat` and an emoticon is matched to be removed but there is no previous node.

Fixes #933

### DIFF
--- a/src/lib/emoticons.js
+++ b/src/lib/emoticons.js
@@ -32,7 +32,7 @@ export function checkWhitespace(node, rangeHelper) {
 		var range = rangeHelper.cloneSelected();
 		var rangeStart = -1;
 		var rangeStartContainer = range.startContainer;
-		var previousText = prev.nodeValue || '';
+		var previousText = (prev && prev.nodeValue) || '';
 
 		previousText += dom.data(emoticon, 'sceditor-emoticon');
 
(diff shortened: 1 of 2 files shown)


src/lib/emoticons.js
```diff
@@ -32,7 +32,7 @@ export function checkWhitespace(node, rangeHelper) {
 		var range = rangeHelper.cloneSelected();
 		var rangeStart = -1;
 		var rangeStartContainer = range.startContainer;
-		var previousText = prev.nodeValue || '';
+		var previousText = (prev && prev.nodeValue) || '';
```
samclarke, line 35: It might be worth doing a breaking change to up the target JS, it's getting annoying not being able to use features like optional chaining plus it would be nice to be able to tidy up the code.

live627 (approved):

PR #947, opened 2023-11-03
Emoji toolbox tool

juhojama:
This PR adds optional unicode emoji selector in toolbar. The existing emoticon support may not be fit for all use cases so now the user has option to use unicode emojies or image emoticons.

Use it by adding preferred emojies in SCEditor initialization:
```
sceditor.create(textarea, {
	format: 'bbcode',
	style: 'https://cdn.jsdelivr.net/npm/sceditor@3/minified/themes/content/default.min.css',
	emojies: ['😊', '❤️', '😄']
});
```
image


samclarke:
Thanks for the PR, it looks really good!

It might be better to have it as a plugin instead of in the default code as not everyone wants emoticons/emoji. I know the emoticon code isn't in a plugin but that's stuck there due to backwards compatibility otherwise I'd move that too.

If you could move it into a plugin I'll get it merged ASAP.

juhojama:
Thanks @samclarke. I was unable to replicate the plugin implementation on my end, but feel free to use the changes in the PR  yourself to make a plugin adaption.

w8tcha:
@juhojama  i created a new PR based on your work but as a plugin implementation.

https://github.com/samclarke/SCEditor/pull/956

PR #956, opened 2024-03-21
Unicode Emoji Toolbox Plugin

w8tcha:
This PR is based on 

https://github.com/samclarke/SCEditor/pull/947

> This PR adds optional unicode emoji selector in toolbar as a Plugin. The existing emoticon support may not be fit for all use cases so now the user has option to use unicode emojies or image emoticons.
> 
> Use it by adding preferred emojies in SCEditor initialization:

```JS
sceditor.create(textarea, {
	format: 'bbcode',
	style: 'https://cdn.jsdelivr.net/npm/sceditor@3/minified/themes/content/default.min.css',
	emojis: ['😊', '❤️', '😄'],
        plugins: 'emojis'
});
```

280264010-358549f5-4611-4409-9514-768764366330

If there are no options set it loads the [EmojiMart](https://missiveapp.com/open/emoji-mart) Picker

emojiMart


src/plugins/emojis.js
```diff
@@ -0,0 +1,86 @@
...
+         */
+        var emojisCmd = function(caller) {
+            const editor = this,
+                content = document.createElement('div'),
+                emojis = editor.opts.emojis || [],
+                perLine = Math.sqrt(Object.keys(emojis).length);
+
+            if (!emojis.length) {
+                const pickerOptions = { onEmojiSelect: handleSelect };
+                const picker = new EmojiMart.Picker(pickerOptions);
```
samclarke, line 34: What is EmojiMart? This will cause an error if it isn't defined. If it's falling back to an external library it might be worth adding a check to make sure it's included.

samclarke (approved):

samclarke:
Thanks for the PR, looks great! :+1: 

The style tests were failing so I've added a commit to fix that as it's just changing spaces to tabs and adding a few extra spaces. Also adds a comment to tell ESLint that `EmojiMart` is a global variable.

Really need to change to using Prettier for the formatting so it can all be automated.

w8tcha:
Thanks for adding it. I also send a pull request for the website documentation

PR #964, opened 2024-05-08
Bind missing events

live627:
The supported events are: 
- 'keyup`
- 'keydown`
- 'Keypress`
- 'blur`
- 'focus`
- 'input`
- 'nodechanged` - When the current node containing the selection changes in WYSIWYG mode
- 'contextmenu`
- 'selectionchanged`
- 'valuechanged`

### What this PR does
Adds callable events that were only accessible via "signals"

-  `pasteraw`
- `paste`
- `pastehtml`

### Goal

To eventually replace signals with events. I believe the concept of event listeners is easier to grasp since everything in the JS   world uses them.

### Question

Should we go one step further and make `init()` and `destroy()` events too?

### DIFF
--- a/src/lib/SCEditor.js
+++ b/src/lib/SCEditor.js
@@ -333,6 +333,7 @@ export default function SCEditor(original, userOptions) {
 		handleMouseDown,
 		handleComposition,
 		handleEvent,
+		handleManualEvent,
 		handleDocumentClick,
 		updateToolBar,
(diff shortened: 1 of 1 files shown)


live627:
Just noticed that those "missing" events are bound to the editor's containing div. Is that a roundabout way to use `CustomEvent` from the DOM?

PR #992, opened 2025-09-27
Buffer node operations when searching for extra whitespace to remove

live627:
This PR optimizes `removeWhiteSpace()` to **defer DOM mutations** instead of removing/modifying nodes immediately. Uses `WeakSet` and `WeakMap` to track nodes for removal and modified values, then applies changes downstream.
- **Problem**: Directly removing nodes in a loop causes repeated DOM reflows
- **Solution**: Buffer node removals and text updates, apply them when converting to BBCode

- **Zero DOM mutations** during whitespace processing — major performance win
- **Backward compatible** — parameters are optional; old code still works
- **Clever buffering** — uses WeakSet/WeakMap to avoid reference leaks

### DIFF
--- a/src/formats/bbcode.js
+++ b/src/formats/bbcode.js
@@ -2250,11 +2250,11 @@
 		 * @private
 		 */
 		function handleBlockNewlines(element, content) {
-			var tag = element.nodeName.toLowerCase();
+			var nodeName = element.nodeName;
 			var isInline = dom.isInline;
 
 			// Only apply to block-level elements or <br>
-			if (!isInline(element, true) || tag === 'br') {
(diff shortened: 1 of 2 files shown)


live627:
My performance profile reveals that writing all these nodes is costly on a large document when toggling modes. 

I was able to get it down to one failing test. I'm going to need help solving that one @samclarke .

src/formats/bbcode.js
```diff
@@ -2241,63 +2241,89 @@
...
+					isInline(previousSibling, true)
+				) {
 					content = '\n' + content;
 				}
 			}
 
 			return content;
 		}
 
+
```
live627:
```suggestion
```

